PR #17, opened 2020-11-25
add python script compilation with pyinstaller in install step

jmkerloch:
- add `pyinstaller `search in `CMakeList.txt`
- if `pyinstaller `found, compile python scripts, otherwise copy pythons scripts in installer
- update GitHub action CI to install `pyinstaller`
- add `requirements.txt` for python scripts dependency (for now only `pyinstaller`)


.github/workflows/dockerHub.yml
```diff
@@ -22,6 +22,15 @@ jobs:
...
     - uses: actions/checkout@v2
     
+    - name: Set up Python
+      run: |
+          apt-get install python3-pip -y
+                    
+    - name: Install dependencies
+      run: |
+          python3 -m pip install --upgrade pip
+          pip install -r src/src_python/requirements.txt
```
flomnes, line 32: pip3

.github/workflows/releases.yml
```diff
@@ -24,6 +24,16 @@ jobs:
     steps:
     - uses: actions/checkout@v2 
     
+    - name: Set up Python
+      uses: actions/setup-python@v1
+      with:
+        python-version: 3.6
```
flomnes, line 30: python 3.6 no longer receives support. Use 3.8 or 3.9 ?

[table of versions](https://en.wikipedia.org/wiki/History_of_Python#Table_of_versions)

flomnes (changes requested):

.github/workflows/dockerHub.yml
jmkerloch: ok changed

.github/workflows/releases.yml
jmkerloch: use 3.8

sgatto (approved):
OK pour moi

.github/workflows/releases.yml
sgatto: ok

flomnes (approved):

PR #18, opened 2020-11-26
Add config.yaml file support

jmkerloch:
Add config.yaml generation from CMake.
Application name are defined from : 
- `$<TARGET_FILE_NAME:exe>` for xpansion executable
- defined variable `ANTARES_SOLVER_APP `for antares-solver 

For antares-solver executable name we need to wait for some work on antares-solver so we can get file name with 
`find_package(antares-solver)` and after that ask for some properties.

For now there is no check on defined values. We must add a parameters check before launch. This should be done in another PR.


src/CMakeLists.txt
```diff
@@ -1,3 +1,17 @@
...
 
-install(DIRECTORY "${CMAKE_CURRENT_SOURCE_DIR}/src_python/" 		DESTINATION xpansion-launcher)
\ No newline at end of file
+#generate config.yaml with executable names in current source directory
+SET(ANTARES_VERSION 7.2)
+SET(ANTARES_SOLVER_APP "antares-${ANTARES_VERSION}-solver")
+
+if (MSVC)
+    SET(ANTARES_SOLVER_APP "${ANTARES_SOLVER_APP}.exe")
+endif()
```
flomnes, line 9: I suggest

```
if (MSVC)
    SET(ANTARES_SOLVER_APP "antares-${ANTARES_VERSION}-solver.exe")
elseif(UNIX)
    SET(ANTARES_SOLVER_APP "antares-${ANTARES_VERSION}-solver")
endif()

```

src/src_python/antares_xpansion/xpansionConfig.py
```diff
@@ -25,7 +40,6 @@ def __init__(self):
```
flomnes, line 40: Why not set `self.MPI_LAUNCHER` and `self.MPI_N` in CMake too ?

src/src_python/antares_xpansion/xpansionConfig.py
```diff
@@ -13,7 +16,19 @@ class XpansionConfig():
...
+    def __init__(self, configfile: Path):
+
+        with open(configfile) as file:
+            content = yaml.full_load(file)
+            if content is not None:
+                self.ANTARES = content.get('ANTARES', "antares-solver")
+                self.MERGE_MPS = content.get('MERGE_MPS', "merge_mps")
+                self.BENDERS_MPI = content.get('BENDERS_MPI', "bender_mpi")
+                self.BENDERS_SEQUENTIAL = content.get('BENDERS_SEQUENTIAL', "benders_sequential")
+                self.LP_NAMER = content.get('LP_NAMER', "lp_namer")
```
flomnes, line 28: Use `sys.platform` to set default values for Linux/Apple "application" or for Windows "application.exe".

So that default values are relevant for each OS family.

src/src_python/antares_xpansion/xpansionConfig.py
jmkerloch: Was done with `sys.platform` code untouched.
I don't think cmake can give us the mpi executable

src/src_python/antares_xpansion/xpansionConfig.py
jmkerloch: I think this will make code less readable and you should have correct values in .yaml because file is generated by cmake.

src/CMakeLists.txt
jmkerloch: OK done

cmake/dependencies/CMakeLists.txt
```diff
@@ -1,58 +1,58 @@
...
-  set(ANTARES_BUILD_TYPE "debug")
-else ()
-  
-  set(ANTARES_BUILD_TYPE "release")
-endif ()
+if (USE_ORTOOLS_STABLE)
+	set(REPOSITORY "https://github.com/google/or-tools.git")
+	set(TAG "stable")
+	set(CMAKE_ARGS "")
+endif()
```
flomnes, line 14: if ... else ... endif

cmake/dependencies/CMakeLists.txt
```diff
@@ -1,58 +1,58 @@
 #####################
-##  Antares solver  ##
+##  ORTOOLS  ##
```
flomnes, line 2: OR-TOOLS

cmake/dependencies/CMakeLists.txt
```diff
@@ -1,58 +1,58 @@
...
+if (USE_ORTOOLS_STABLE)
+	set(REPOSITORY "https://github.com/google/or-tools.git")
+	set(TAG "stable")
+	set(CMAKE_ARGS "")
+endif()
 
 build_git_dependency(
     NAME
-      antares-solver
+      ortools
```
flomnes, line 18: or-tools

src/src_python/antares_xpansion/xpansionConfig.py
flomnes: Then remove default values and throw an error if any attribute is not found (future revision)

flomnes (changes requested):

flomnes (approved):

PR #19, opened 2020-12-02
add arguments for number of MPI process

jmkerloch:
- Move arguments parsing from config.py to driver.py.
- Add number of MPI process argument

### DIFF
--- a/src/src_python/antares_xpansion/driver.py
+++ b/src/src_python/antares_xpansion/driver.py
@@ -8,7 +8,6 @@ import glob
 import os
 import subprocess
 import sys
-import argparse
 
 from pathlib import Path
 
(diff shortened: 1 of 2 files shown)


src/src_python/antares_xpansion/driver.py
```diff
@@ -29,7 +30,38 @@ def __init__(self, config):
...
+                            required=True)
+        parser.add_argument("--method",
+                            dest="method",
+                            type=str,
+                            choices=["mpibenders", "mergeMPS", "both", "sequential"],
+                            help="Choose the optimization method")
+        parser.add_argument("-c",
+                            dest="c",
+                            help='Name of the file to use for exclusion constraints')
+        parser.add_argument("-n", "-np",
```
sgatto, line 59: je ne suis pas sur que `-np` soit accepté par add_argument, normalement un `--` est nécessaire

sgatto (commented):
I'm not convinced to give the responsibility of parsing the argument to `driver` instad of `xpansionConfig.py`. I'd rather give the responsibility to xpansionConfig instead of driver.
I can propose the changes

src/src_python/antares_xpansion/driver.py
jmkerloch: En effet. Remarques prises en compte.

jmkerloch:
> 
> 
> I'm not convinced to give the responsibility of parsing the argument to `driver` instad of `xpansionConfig.py`. I'd rather give the responsibility to xpansionConfig instead of driver.
> I can propose the changes

I agree, changes done. `xPansionConfig `now store the configuration parsed.

Added also a check on N_MPI parameter.

sgatto (approved):
OK

PR #38, opened 2021-02-02
Feature/example test

jmkerloch:
Add some pytest to check antares-xpansion result.
Log are checked for investment solution and last iteration objective function values.


src/tests/CMakeLists.txt
```diff
@@ -0,0 +1,25 @@
+cmake_minimum_required(VERSION 3.10)
+
+project(unit_tests_antares_xpansion NONE )
+
+find_package(Python3 EXACT 3.8 COMPONENTS Interpreter)
+if(Python3_Interpreter_FOUND)
```
flomnes, line 6: This condition is also checked in function `find_python_module`, consider removing one check

src/tests/examples/example_test.py
```diff
@@ -0,0 +1,106 @@
...
+                    investment_solution[investment_solution_list[0].strip()] = float(investment_solution_list[1])
+                res = log_file.readline().strip()
+        log = log_file.readline()
+
+    log_file.close()
+
+    np.testing.assert_allclose(last_it[1:3], expected_last_it[0:2], rtol=1e-4, atol=0)
+
+    for investment in expected_investment_solution.keys():
+        assert investment in investment_solution.keys(), "Investment " + investment + " not found in solution"
```
flomnes, line 87: Is this assertion needed ? It seems that it is always verified, looking at the for loop above

src/tests/examples/example_test.py
```diff
@@ -0,0 +1,106 @@
...
+    study_path = ALL_STUDIES_PATH / "xpansion-test-01"
+
+    expected_last_it = [2.2738256189e+10, 2.2738256189e+10]
+    expected_investment_solution = {"battery" : 1.0e+03, "peak" : 1.4e+03 , "pv" : 1.0e+03 , "semibase" : 2.0e+02}
+    
+    launch_xpansion(installDir, study_path, "sequential")
+    check_investment_solution(study_path,expected_last_it,expected_investment_solution)
+    
+    launch_xpansion(installDir, study_path, "mpibenders")
+    check_investment_solution(study_path, expected_last_it,expected_investment_solution)
```
flomnes, line 104: Consider splitting into two tests "sequential" and "mpibenders"

src/tests/examples/example_test.py
```diff
@@ -0,0 +1,106 @@
...
+                if ("=" in res):
+                    investment_solution_list = res.replace("|", "").strip().split("=")
+                    assert (len(investment_solution_list) == 2)
+                    investment_solution[investment_solution_list[0].strip()] = float(investment_solution_list[1])
+                res = log_file.readline().strip()
+        log = log_file.readline()
+
+    log_file.close()
+
+    np.testing.assert_allclose(last_it[1:3], expected_last_it[0:2], rtol=1e-4, atol=0)
```
flomnes, line 84: Add a TODO note to use results instead of logs (if possible)

flomnes (changes requested):

src/tests/examples/example_test.py
jmkerloch: We need to check if the investment for a link is present in the result file.

In the loop abose we are populating the investment_solution map but not checking if the link was expected.

flomnes: So you want to check that all elements of `expected_investment_solution.keys()` are in `investment_solution.keys()`, OK

flomnes (approved):

PR #49, opened 2021-02-24
Use of precompiled antares-solver

### LINKED ISSUE #44: Avoid submodule init when compiling antares-solver
When compiling antares-solver, all submodules are used : 
- antares-deps
- examples
- tests

We need to define `GIT_SUBMODULES `CMake `ExternalProject_Add `option to init only antares-deps.

jmkerloch:
Use of antares-solver 8.0.0 and download of pre compiled archive for CI

closes #44 : only antares-deps submodule is used for antares-solver compilation


flomnes (commented):
It would be great if the solver was a build asset. But it would require to embed all dynamic libs into the executable, I don't know if we want to do this.

flomnes (commented):
If you want to use Antares Simulator 9 when it will be released, you will have to replace all 8.0.0 by 9.0.0 in this commit.

You could create a .json file similar to this
```json
{
	"antares_version": "8.0.0"
}
```
and use [this GH Action](https://github.com/marketplace/actions/get-json-property) to read the version number from this file.

When the version changes, you will only have to change this file.

Issue #45, opened 2021-02-23
Error when using the Antares "playlist" function


MathildeFranconRTE:
[xpansion-test-02.zip](https://github.com/AntaresSimulatorTeam/antares-xpansion/files/6034878/xpansion-test-02.zip)

The use of the "MC scenario playlist" function to run Antares over 3 MC years out of the 10 MC years did not work on the test_case study (but it worked on a previous version of the open-source version of Xpansion) 

Traceback (most recent call last):
  File "launch.py", line 11, in <module>
  File "antares_xpansion\driver.py", line 219, in launch
  File "antares_xpansion\driver.py", line 483, in generate_mps_files
  File "antares_xpansion\driver.py", line 287, in pre_antares
  File "configparser.py", line 697, in read
  File "configparser.py", line 1093, in _read
configparser.DuplicateOptionError: While reading from 'E:\\AppliRTE\\Documents\\antaresXpansion\\Opensource\\antaresXpansion-0.1.0-win64\\antaresXpansion-0.1.0-win64\\examples\\xpansion-test-01\\settings\\generaldata.ini' [line 75]: **option 'playlist_year +' in section 'playlist' already exists**
[11772] Failed to execute script launch

jmkerloch:
python module `configparser` is used to parse .ini files.

By default, if there are multiple section with same name a `DuplicateOptionError `is thrown.

jmkerloch:
There is also an issue with `antares-solver` in case of playlist use with `antares-xpansion` : 

```
python launch.py --step full --dataDir ../../../xpansion-test-02
['pv', 'battery', 'semibase1', 'semibase2', 'peak1', 'peak2']
starting mps generation
-- pre antares
-- launching antares
['C:\\Projet_SRO\\rte-antares\\antares-xpansion-installer-64bits\\antaresXpansion-0.1.0-win64\\bin\\antares-8.0-solver.exe', 'C:\\Projet_SRO\\rte-antares\\xpansion-test-02']
['20210224-1534eco', 'maps']
['20210224-1534eco', '20210224-1542eco', 'maps']
-- post antares
Traceback (most recent call last):
  File "launch.py", line 12, in <module>
    DRIVER.launch()
  File "C:\Projet_SRO\rte-antares\antares-xpansion-installer-64bits\antaresXpansion-0.1.0-win64\xpansion-launcher\antares_xpansion\driver.py", line 219, in launch
    lp_path = self.generate_mps_files()
[...]

PR #50, opened 2021-02-24
Fix : parse ini file with duplicate sections

jmkerloch:
See #45 

To parse `generaldata.ini` with multiple `playlist +` section, we must use `strict = False` option for `ConfigParser `use

### DIFF
--- a/src/src_python/antares_xpansion/driver.py
+++ b/src/src_python/antares_xpansion/driver.py
@@ -205,7 +205,7 @@ class XpansionDriver():
         """
             returns the nubyears parameter value read from the general data file
         """
-        ini_file = configparser.ConfigParser()
+        ini_file = configparser.ConfigParser(strict=False)
         ini_file.read(self.general_data())
         return float(ini_file['general']['nbyears'])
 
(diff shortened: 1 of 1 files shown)


flomnes (changes requested):
Here is an excerpt of a problematic generaldata.ini configuration file, which we'll call **playlist.ini**.

```ini
[playlist]
playlist_reset = false
playlist_year + = 0
playlist_year + = 5
playlist_year + = 8
```

It is problematic because it has a duplicate item `playlist_year`. It is my opinion that we should avoid duplicate items/sections in .ini files, because they are not well supported by module ConfigParser. As shown below, only the last value is taken into account.

```python3
import configparser as cp
parser=cp.ConfigParser(strict=False)
parser.read_file(open('playlist.ini'))
print(dict(parser['playlist']))
# Result
{'playlist_reset': 'false', 'playlist_year +': '8'}
```
To me, a better solution would be to use real lists like this
```ini
[playlist]
playlist_reset = false
playlist_year = [0,5,8]
```

```python3
import configparser as cp

def string_to_int_list(s : str):
    """
    Convert a string representation of a list of ints 
    into a list of ints
    
    Example :
    >>> string_to_int_list('[1,2]')
    >>> [1,2]
    """
    return [int(x) for x in s.strip('[ ]').split(',')]

parser=cp.ConfigParser(strict=False)
parser.read_file(open('playlist_list.ini'))
print(string_to_int_list(parser.get('playlist', 'playlist_year')))
# Result
[0, 5, 8]
```

It would require to change Antares Simulator's reader a bit to support this new format.

flomnes:
See [this PR](https://github.com/AntaresSimulatorTeam/Antares_Simulator/pull/181)

flomnes (approved):
Ok, playlist is not used by xpansion.

PR #64, opened 2021-03-08
Correction bug JSON operational_cost

Blaxav:
Correcting bug : operational_cost = xxe-312 in out.json with bendersmpi.
The operational cost was not computed in the main of bendersmpi.


flomnes (commented):
Is mini_network a new test-case ? If so, how to launch it ?

Blaxav:
It is, but I saw a bug on it doing the tests. I will push the corrected one with the tests.
It's a test case at the optimization level (after lpnamer). To launch this test, and the other mini_instances, you can use the command : 
intallDir/benderssequential.exe options_default.txt
or
mpiexec -n N installDir/bendersmpi.exe options_default.txt
from the instance folder, where N is the number of wanted mpi nodes.
I will write those commands in a README.

jmkerloch:
See issue #66 for benders data_test use.

PR #100, opened 2021-04-02
Features/end to end tests update study

### LINKED ISSUE #97: Update Antares study with the optimal solution at the end of the optimization
Here is a description of the latest feature implemented by Eurodecision for you to test.

Once antaresXpansion optimisation is complete, the aim is to update the Antares capacity times-series with the antaresXpansion solution in the Antares study.

1. The optimal capacity _x_ of an Xpansion candidate is a number in MW: its value can be found for instance at the end of the out.json file linked to the candidate name (e.g. battery = 997 )

2. The rest of the information on each candidate is found in the  /user/expansion/candidates.ini file:

    - _**name**_ : the name of the candidate

    - _**link**_ : the link name of the corresponding candidate in Antares. For example, if this link connects the nodes "area1" and "area2", the link name will be "area1 - area2";

    - _**link-profile**_: name of the text file in user/expansion/capa/, containing 1 or 2 columns (2 columns if you want to differentiate between direct and indirect, 1 column if the capacity is the same in direct and indirect) of 8760 rows of numbers between 0 and 1 representing the percentage of the installed capacity actually available at each hour of the 8760 hours of the year;

    - _**already-installed-capacity**_ and possibly _**already-installed-link-profile**_: if the candidate already has installed capacity before optimisation, and possibly a link-profile applies to this already installed capacity (note: in this case link-profile and already-installed-link-profile must necessary have the same format);

    - _**has-link-profile**_ = True: indicates that there is a link-profile to take into account for this candidate.

The following formula defines the link between :
- the optimal installed capacity x by antaresXpansion, 
- the parameters of the candidate,
- the capacity actually available in the Antares study, hour by hour and in both directions of the link concerned:
image

- The Antares time-series capacities correspond to 2 columns (direct and indirect direction) of 8760 lines, and are found in the first 2 columns of the Antares link which is in study\input\links\area1\area2.txt (with our example area1 and area2), which has 8 columns in all (or less ....depends on the Antares version,) 

- If the candidate has no link-profile: x (or possibly x + x0) is copied and repeated in both columns 8760 times.

- If the candidate has a link-profile with only 1 column: the column is simply duplicated (the 2nd column is equal to the 1st).

Example: if I invest 500 MW in the candidate grid_a1_a2 with Xpansion :

image

Update Antares study like that : 
image

Last thing: 
For the moment, we can only have 1 candidate per link, but I would like to have several in the future.
If you do a test, it would be good to anticipate this: in this case, you would have to **sum the hourly capacities obtained for each candidate of the same link**.

Classic test-case with candidates links set to 0 in Antares at beginning : 
[xpansion-test-01-update-study.zip](https://github.com/AntaresSimulatorTeam/antares-xpansion/files/6223617/xpansion-test-01-update-study.zip)

jmkerloch:
closes #97


src/tests/examples/example_test.py
```diff
@@ -51,6 +53,8 @@ def verify_solution(study_path, expected_values, expected_investment_solution):
 
     solution = json_data["solution"]
     investment_solution = solution["values"]
+    
+    json_file.close()
```
sgatto, line 57: we could use 
```python
with open(str(json_path), 'r') as json_file:
    json_data = json.load(json_file)
```
wthout the need to `close` the file

sgatto (approved):

src/tests/examples/example_test.py
jmkerloch: OK but what will happen if there is an assert raised ?

I added the close statement because in Windows we couldn't launch another test because json output file was already used.

PR #104, opened 2021-04-06
multiple xpansion launch

### LINKED ISSUE #99: antares-xpansion can't be launched twice at the same
We can't launch 2 antares-xpansion simulation at the same time when using a single antares-xpansion installed version : 
```
PS C:\Projet_SRO\rte-antares\antares-xpansion\src\src_python> python launch.py --installDir ..\..\_buildRelease\Release\ --dataDir ..\..\examples\xpansion-test-02-new\ --step full --method mpibenders
Traceback (most recent call last):
  File "launch.py", line 12, in <module>
    DRIVER.launch()
  File "C:\Projet_SRO\rte-antares\antares-xpansion\src\src_python\antares_xpansion\driver.py", line 189, in launch
    self.clear_old_log()
  File "C:\Projet_SRO\rte-antares\antares-xpansion\src\src_python\antares_xpansion\driver.py", line 235, in clear_old_log
    os.remove(self.exe_path(self.config.LP_NAMER) + '.log')
PermissionError: [WinError 32] Le processus ne peut pas accéder au fichier car ce fichier est utilisé par un autre processus: 'C:\\Projet_SRO\\rte-antares\\antares-xpansion\\_buildRelease\\Release\\lp_namer.exe.log'
```

We need to use a specific log file for each simulation.

jmkerloch:
change lp_namer and study_updater log path to output directory for multiple xpansion launch
closes #99 

There is still an issue if the same study is launched at the same time.
In this case we can't define which antares-solver output directory is the current directory and an assert is raised : 

```
new_output = os.listdir(self.antares_output())
assert len(old_output) + 1 == len(new_output)
diff = list(set(new_output) - set(old_output))
```

### DIFF
--- a/src/src_python/antares_xpansion/driver.py
+++ b/src/src_python/antares_xpansion/driver.py
@@ -297,19 +297,15 @@ class XpansionDriver():
         # if not os.path.isdir(driver.antares_output()):
         #     os.mkdir(driver.antares_output(), )
         old_output = os.listdir(self.antares_output())
-        with open(self.get_antares_log_filename(), 'w') as output_file:
-            returned_l = subprocess.run(self.get_antares_cmd(), shell=False,
-                                        stdout=output_file,
-                                        stderr=output_file)
-            if returned_l.returncode != 0:
-                print("WARNING: exited antares with status %d" % returned_l.returncode)
+        returned_l = subprocess.run(self.get_antares_cmd(), shell=False,
+                                        stdout= subprocess.DEVNULL,
+                                        stderr= subprocess.DEVNULL)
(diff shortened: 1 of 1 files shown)


flomnes (changes requested):
We might need antares output for debugging purposes. So it is not desirable to redirect it to `/dev/null`.

jmkerloch:
> 
> 
> We might need antares output for debugging purposes. So it is not desirable to redirect it to `/dev/null`.

Even if redirection is done to `/dev/null` we still have antares solver log in antares output directory

flomnes (approved):

PR #117, opened 2021-04-21
Features/logger

### LINKED ISSUE #93: create logger module for benders
Currently `std::cout` is used to display some information to user during optimisation.

We need to introduce a logger module so `std::cout` can display different information if we are using antares-xpansion as a user or as a developper.

This logger should also be able to add log to a log file.

jmkerloch:
related #93


flomnes (commented):
Why not use inheritance ?
```cpp
struct LogData {
// Logged attributes
};

struct BendersData : public LogData {
// Non-logged attributtes
};

void log_at_initialization(const LogData& d) {
// ...
}

// Call
BendersData bd;
log_at_initialization(bd); // OK: BendersData derives from LogData
```
So you don't need to copy duplicate attributes like here
```cpp
LogData bendersDataToLogData(const BendersData &data) {
    LogData result;
    result.lb = data.lb;
    result.best_ub = data.best_ub;
    result.it = data.it;
    result.slave_cost = data.slave_cost;
    result.invest_cost = data.invest_cost;
    result.x0 = data.x0;
    result.min_invest = data.min_invest;
    result.max_invest = data.max_invest;
    return result;
}
```

sgatto:
We have to think about the architecture a little better. We might fall in a circular dependency problem

PR #127, opened 2021-05-03
Change pyinstaller: 2 options, single file and  multiple binaries

sgatto:


jmkerloch (approved):
User documentation must be updated before release with new install folder refactoring.

PR #152, opened 2021-05-20
Feature/multi solver interface

jmkerloch:
The PR is huge as the interface modifies many parts of the code.
Maybe the focus can be first done on src/cpp/multisolver_interface/ files

Still requires some changes :
  * unit-tests need to be modified to integrate the interface
  * Some end-to-end tests don't pass : there are still some bugs in lp_namer integration. Some of the tests don't pass because of numerical differences, this need further investigations but might happen because some optimal solutions of problems are solver dependant.

Already done :
   * Interface with COIN, CPLEX and XPRESS
   * Coin has been integrated with the distribution provided by ORTools, so that the code generation is the same as before
   * benders_sequential, benders_mpi and merge_mps work
   * interface tests are still with catch2, i didn't change it for now
  
Also added some comments in lp_namer as I think some things are useless


src/cpp/benders_sequential/Worker.cpp
```diff
@@ -36,37 +43,41 @@ void Worker::get_value(double & lb) {
...
-		else
-		{
-			error_l = true;
-			std::cout << "\nERROR : missing variable " << kvp.first << " in " << path_to_mps;
+		else {
+			_solver = factory.create_solver(solver_name);
 		}
+		_solver->init();
+	}
+	_solver->set_threads(1);
```
jmkerloch, line 72: Why do we need to use only one thread ?

PR #196, opened 2021-07-26
Fix/run vcpkg

jmkerloch:
- use of vcpkg as submodule instead of github action run-vcpkg
- use of github action to cache vcpkg installed package
- test of ccache use to speed up compilation on ubuntu


sgatto (approved):
I like it

PR #212, opened 2021-08-17
Feature/several investment candidates 07 study updator

Sefra35S:
Update of study updator to support multiple candidates on the same link.


jmkerloch (approved):
OK for me

PR #216, opened 2021-08-23
WIP : MasterProblemBuilder

Sefra35S:


src/cpp/exe/lpnamer/main.cpp
```diff
@@ -63,110 +64,50 @@ std::string get_name(std::string const & path) {
...
-		}
-		++i;
-	}
+	std::vector<Candidate> candidates;
 
-    solver_addcols(master_l, obj_interco, mstart, {}, {}, lb_interco, ub_interco, coltypes_interco, interco_names);
-
-	// integer constraints
-	int n_integer = pallier.size();
-	if(n_integer>0 && master_formulation=="integer"){
```
jmkerloch, line 115: `master_formulation` parameter is no longer use. There is a risk of regression

src/cpp/lpnamer/MasterProblemBuilder.cpp
```diff
@@ -0,0 +1,129 @@
...
+	addVariablesPmaxOnEachCandidate(candidates, master_l);
+
+	std::vector<Candidate> candidatesInteger;
+	for (int i = 0; i < candidates.size(); i++)
+	{
+		const auto& candidate = candidates.at(i);
+		if (candidate.is_integer()) {
+			candidatesInteger.push_back(candidate);
+		}
+	}
```
jmkerloch, line 25: Maybe we could use `std::copy_if` ?

src/cpp/lpnamer/MasterProblemBuilder.cpp
```diff
@@ -0,0 +1,129 @@
...
+
+	addNvarOnEachIntegerCandidate(candidatesInteger, master_l);
+
+	addPmaxConstraint(candidatesInteger, master_l);
```
jmkerloch, line 29: There should be a `if` depending on `master_formulation`

jmkerloch (changes requested):
There is a risk of regression because `master_formulation` parameter is no longer use.

PR #226, opened 2021-09-08
Fix/direct indirect hurdles cost

### LINKED ISSUE #182: Bug when hurdle costs are activated on candidate link
The problem is related to additional constraints added in Antares when there are hurdle costs on the candidate link.

•	Indeed, if the "Use hurdles cost" button is green, those additional constraints  appear in Antares (even if hurdle costs are set to zero …).
image

To detail a little more mathematically, p[t] is the flow on the candidate link and Pmax is the capacity of the link. 
Here are the constraints that are in Antares :

-       p[t] reprensents the flow on the link

- 	–pMax_indirect <= p[t] <= pMax_direct, --> Xpansion changes this Pmax value with each iteration, taking into account already_installed_capacity and new investments

- 	 p[t]=p+[t] - p-[t] and p+[t]≤pMax_direct and p-[t]≤pMax_indirect , --> those  last constraints appear only when there are hurdle costs and Xpansion has not updated them with the right Pmax, this is a bug that needs to be fixed. Those constraints take into account the Pmax capacities that are in the Antares interface instead of taking the capacities of Xpansion, which can limit the Xpansion solution. In order not to affect the optimization and avoid the bug, you do as you did: put a large non-limiting capacity (e.g. 10000 MW, and use an already_installed_capacity in the candidate.ini file if you want, as you did), and you can also put positive hurdle costs if you want. This is a quick fix until we fix the tool.

Name of variables  p+ et p- in Antares :

CoutOrigineVersExtremiteDeLInterconnexion
CoutExtremiteVersOrigineDeLInterconnexion

Modification of MPS problem file for each hour (here hour = 141) of the week :
 UP BOUND                                             CoutOrigineVersExtremiteDeLInterconnexion_0_141_       2000.01
 UP BOUND                                             CoutExtremiteVersOrigineDeLInterconnexion_0_141_       2000.01

--> becomes :
 UP BOUND                                             CoutOrigineVersExtremiteDeLInterconnexion_0_141_       Xpansion_capacity_variable
 UP BOUND                                             CoutExtremiteVersOrigineDeLInterconnexion_0_141_       Xpansion_capacity_variable

jmkerloch:
closes #182

### DIFF
--- a/tests/cpp/lp_namer/ProblemModifierTest.cpp
+++ b/tests/cpp/lp_namer/ProblemModifierTest.cpp
@@ -9,6 +9,10 @@
 const std::string P_LINK = "p_link";
 const std::string P_PLUS = "p_plus";
 const std::string P_MINUS = "p_minus";
+const double ZERO = 0.0;
+const double PLUS_INF = PLUS_INF;
+const double MINUS_INF = MINUS_INF;
+
 
 class ProblemModifierTest : public ::testing::Test
 {
@@ -252,9 +256,9 @@ TEST_F(ProblemModifierTest, One_link_no_candidates_link_boundaries_are_removed) 
 
     verify_columns_are(3);
 
-    verify_column(0,P_LINK,'C',1,-1e20,1e20);
-    verify_column(1,P_PLUS,'C',1,0,1e20);
-    verify_column(2,P_MINUS,'C',1,0,1e20);
+    verify_column(0,P_LINK,'C',1,MINUS_INF, PLUS_INF);
+    verify_column(1,P_PLUS,'C',1,ZERO, PLUS_INF);
+    verify_column(2,P_MINUS,'C',1,ZERO, PLUS_INF);
 
     try {
         problem_modifier.get_candidate_col_id("invalid_cand_name");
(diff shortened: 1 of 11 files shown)


tests/cpp/lp_namer/ProblemModifierTest.cpp
```diff
@@ -319,121 +355,159 @@ TEST_F(ProblemModifierTest, One_link_two_candidates_two_timestep) {
...
+    math_problem = problem_modifier.changeProblem(std::move(math_problem), links, p_var_columns,p_direct_cost_columns,p_indirect_cost_columns);
+    verify_columns_are(5);
+
+    const int P_LINK_id  = 0;
+    const int P_PLUS_id = 1;
+    const int P_MINUS_id = 2;
+    const int cand1_id= 3;
+    const int cand2_id= 4;
+
+    verify_column(P_LINK_id,P_LINK,'C',1,-1e20,1e20);
```
sgatto, line 367: we can "rename"
`0` into `ZERO` 
`1e20` into `PLUS_INF` 
`-1e20` into `MINUS_INF` 

the third argument, if I remember well corresponds to the  variable type
1=double (or continuous)
0=integer

we could name two constants:
`INTEGER_VAR`
`CONTINUOUS_VAR`
 Continuous or other, I dontremeber how the multisolver interface colls it

sgatto (changes requested):
only one small observation? not mandatory

tests/cpp/lp_namer/ProblemModifierTest.cpp
jmkerloch: third argument is objective. No new constant defined

sgatto (approved):

PR #291, opened 2021-11-24
Introduce relative gap

tbittar:
- New option `relative_gap` in `settings.ini`
- Correct negative gap display
- Set store in input to false in test examples 
- Update expected output of end-to-end tests due to the previous changes

PR #302, opened 2021-12-01
Fix/broken medium tests

### LINKED ISSUE #300: Medium tests are broken
Medium tests are broken after PR #284. We should check the changes in `example_test.py` done in PR #284 and PR #291.

tbittar:
- Split long tests in sequential and mpi in the same way as medium tests
- Restore the Antares study in its initial state after running the tests
- Check that `optimality_gap` and `relative_gap` match the expected values
- Restore the expected values introduced in PR #291 that were overwritten when merging PR #284 

This PR closes #300


a-zakir:
ok for this pr. 
some 'long' tests failed, further investigation is needed --> see #305

a-zakir (approved):

PR #311, opened 2021-12-07
remove xpress uncontrolled stdout

sgatto:
Tis fix aims to remove the stop Xpress from printing log information on the tdout that concern the reading phase of empty problems.
This fix does remove several prints of this type that precede the messages that `benders`writes for the user
```
Reading Problem empty
Problem Statistics
           0 (      0 spare) rows
           0 (      0 spare) structural columns
           0 (      0 spare) non-zero elements
Global Statistics
           0 entities        0 sets        0 set members

```

This fix does not look for a solution for the solvers logs.
A separate issue should be created


sgatto:
the issue is not fixed

tbittar (approved):

PR #317, opened 2021-12-10
fix: improve a bit the additional constraints' reader logs

sgatto:
a badly written additional constraint file created by the user lead to an error of lp-namer.
The `cout` message was very poor.
This fix add a small information concerning where the problem was raised

### DIFF
--- a/src/cpp/lpnamer/input_reader/AdditionalConstraintsReader.cpp
+++ b/src/cpp/lpnamer/input_reader/AdditionalConstraintsReader.cpp
@@ -37,7 +37,7 @@ void AdditionalConstraintsReader::processSectionLine()
 {
     if ( _line[_line.length()-1] != ']' )
     {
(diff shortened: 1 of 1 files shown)


src/cpp/lpnamer/input_reader/AdditionalConstraintsReader.cpp
```diff
@@ -37,15 +37,15 @@ void AdditionalConstraintsReader::processSectionLine()
 {
     if ( _line[_line.length()-1] != ']' )
     {
-        std::cout << "line " << _lineNb << " : section line not ending with ']'.\n";
+        std::cout << "additional constraints file:  line " << _lineNb << " : section line not ending with ']'.\n";
```
a-zakir, line 40: how about to remind the additional constraints file name --> "in additional constraints file:" _path_to_file<<" line "...

PR #322, opened 2021-12-15
use Path class

### LINKED ISSUE #98: Hide the platform-dependent separator mess inside a class
## Hide the platform-dependent separator mess inside a class
Create a wrapper around a `std::string` to mimic `std::filesystem::path`, without adding the painful C++17 dependency.

```cpp
#include <iostream>
#include <string>

class Path {
private:
  std::string mPath;
#if defined(WIN32) || defined(_WIN32) || defined(__WIN32__) || defined(__NT__)
  // Windows
  static const char mSep = '\\';
#else
  // Linux, Unix, Apple
  static const char mSep = '/';
#endif
public:
  Path(const std::string& s) : mPath(s){}
  Path(std::string&& s) : mPath(std::move(s)){}
  Path operator/(const Path& other) {
    return Path(mPath + mSep + other.mPath);
  }
  Path operator/(const std::string& s) {
    return Path(mPath + mSep + s);
  }
  operator std::string() const {
    return mPath;
  }
  friend std::ostream& operator<<(std::ostream& os, const Path& p) {
    os << p.mPath;
    return os;
  }
};

int main() {
  Path path("directory");
  Path file("file.txt");
  std::cout << path / file << std::endl;
  std::cout << path / "hello.txt" << std::endl;
}
```
## Possible output (Linux)
```
directory/file.txt
directory/hello.txt
```

## Possible output (Windows)
```
directory\file.txt
directory\hello.txt
```

a-zakir:
Closes #98


src/cpp/helpers/include/helpers/Path.h
```diff
@@ -0,0 +1,35 @@
...
+  }
+  Path operator/(const std::string& s) const {
+    return Path(mPath + mSep + s);
+  }
+  explicit operator std::string() const {
+    return mPath;
+  }
+  friend std::ostream& operator<<(std::ostream& os, const Path& p) {
+    os << p.mPath;
+    return os;
```
sgatto, line 43: can we add a `get_string` method that returns a const_str to avoid to explicitly static cast when we need to use a `Path`object?

sgatto (changes requested):
I'm not sure about the need of multiple `static_cast`.
Can we explore to add an explicit method that returns a const_str?

src/cpp/helpers/include/helpers/Path.h
a-zakir: static_cast call `explicit operator std::string()`, but i agree with you, a "get_string" method would be clearer.

sgatto (approved):

sgatto (approved):

PR #334, opened 2021-12-20
Feature/benders output writer

### LINKED ISSUE #327: create output writer interface
in order to harmonize benders outputs, create an (class) interface (OutputWriter) inherited by the actual JsonWriter class.

- the JsonWriter class will be used by the Sequential benders and by the master proc (world.rank =0) of Mpi benders.
-  a "void" class (voidWriter) that write nothing has to derive from OutputWriter if world.rank !=0.

### LINKED ISSUE #347: Incorrect problem status in `out.json`
In the `out.json` file, the field `solution : {problem_status : xxx}` is incorrect when the algorithm stops because of the timelimit. For now it only handles the cases `OPTIMAL`, `MAX_ITERATIONS` and `ERROR`. The case where the algorithm stops with the timelimit criterion is not handled for now and leads to `ERROR`.

We must ensure that all termination criteria of the algorithm are handled when writing the problem status in `out.json`. For the timelimit case, we propose to write `problem_status : "TIMELIMIT"`.

a-zakir:


sgatto:
I think the general idea is very well there.

I think it will bve better if `OutputWriter` is a pure abstract class (only virtual methods) that does not require any test.`

So I propose to name differently the `write` functins of `OutputWriter`, in order to be more specific of the exact functions that they implement and think about their interface based on the use that the `Benders` classes will make.
We can discuss this a bit further.

PR #343, opened 2022-01-05
refactor and add tests for input_checker.py

### LINKED ISSUE #336: refactoring of input_checker.py
`input_checker.py` needs to be refactored to improve its design, redibility and adptability to future changes.
This file is not covered by tests despite its liited scope.

We should refctor and test this portion of the code

a-zakir:
Closes #336


tests/python/test_input_checker.py
```diff
@@ -0,0 +1,321 @@
...
+
+            check_candidate_name("escapeR\r123", "section1236")
+
+            check_candidate_name("tab\t123", "section1236")
+
+            check_candidate_name("escapeF\f\v123", "section1236")
+
+            check_candidate_name("escapeV\v123", "section1236")
+
+            check_candidate_name("parenthesis(123", "section1236")
```
sgatto: with pytest.raises(IllegalCharsInCandidateName):

should be used for each "check"

tests/python/test_input_checker.py
```diff
@@ -0,0 +1,321 @@
...
+            check_candidate_link("", "section0")
+
+    def test_fail_if_link_doesnt_contains_separator(self):
+
+        with pytest.raises(CandidateLinkWithoutSeparator):
+
+            check_candidate_link("atob", "section")
+            check_candidate_link("ato-b", "section")
+            check_candidate_link("ato- b", "section")
+            check_candidate_link("ato -b", "section")
```
sgatto: same here,

with pytest.raises(CandidateLinkWithoutSeparator)

should be used for each check

sgatto (approved):

src/python/antares_xpansion/input_checker.py
```diff
@@ -88,29 +115,28 @@ def check_candidate_option_type(option, value):
                      'already-installed-capacity': 'non-negative',
                      'already-installed-link-profile': 'string',
                      'has-link-profile': 'string'}
-    obsolete_options = ["has-link-profile"]
+    obsolete_options = ["c"]
```
tbittar: In `options_types` there are some old (now unused) options :
- `enable`
- `candidate-type`
- `investment-type`
- `relaxed`
- `has-link-profile`

The thing is that many users still have these options in their `candidates.ini` file and we do not want to raise an exception but only a warning when these options are present.

With the current code, no warning is issued as they are valid `option_types`, we should add these options in `obsolete_options` so that the warning is printed. This is a quick but not elegant solution as we need to repeat the name of unused options. It may be worth to find a way to "tag" options : maybe `"option": (type, deprecated/useful)` similarly as what is done in `options_types_and_legal_values` line 401

Also uniformize the format `option_types` with the `options_types_and_legal_values` used for the settings line 401

By the way, I don't understand why there is a `"c"` in `obsolete_options` right now

src/python/antares_xpansion/input_checker.py
```diff
@@ -88,29 +115,28 @@ def check_candidate_option_type(option, value):
...
-        flushed_print('check_candidate_option_type: Non handled data type %s for option %s'
-                      % (option_type, option))
-        sys.exit(1)
+
+
+class IllegalCandidateOptionValue(Exception):
+    pass
 
 
 def check_candidate_option_value(option, value):
```
tbittar: I don't really understand the purpose of this function.

We only seem to check if options that can be only `true` or `false` are correctly set. No check is performed for the other options. The title is also misleading as I would also expect that we check if options expecting positive floats get a correct value for example (but this is already done in `check_candidate_option_type`).

Moreover, these checks only concern deprecated function, so I think this function can be completely removed.

src/python/antares_xpansion/input_checker.py
```diff
@@ -137,12 +162,20 @@ def check_candidate_option_value(option, value):
...
+
+class EmptyCandidateName(Exception):
+    pass
+
+
+class IllegalCharsInCandidateName(Exception):
+    pass
 
 
 def check_candidate_name(name, section):
```
tbittar: The description may be updated to say that we check that the candidate name is valid instead of "not empty and does not contain space". 

In fact the description could be completely removed but we may want to keep the coherence with the other function of the script that have a description.

src/python/antares_xpansion/input_checker.py
```diff
@@ -168,58 +209,45 @@ def check_candidate_link(link, section):
...
-    # check that link is not empty
-    for each_section in ini_file.sections():
-        check_candidate_name(
-            ini_file[each_section]['name'].strip(), each_section)
-        check_candidate_link(
-            ini_file[each_section]['link'].strip(), each_section)
-        driver.candidates_list.append(
-            ini_file[each_section]['name'].strip().lower())
 
+def check_name_and_link_unicity(ini_file):
```
tbittar: We only check name unicity here, the function should be renamed and the following comment may be removed

(It is indeed possible to have several investment candidates on the same link)

src/python/antares_xpansion/input_checker.py
```diff
@@ -229,11 +257,25 @@ def check_candidates_file(driver):
             if value in unique_values:
                 flushed_print('Error candidates %ss have to be unique, duplicate %s %s in section %s'
                               % (verified_attribute, verified_attribute, value, each_section))
-                sys.exit(1)
+                raise CandidateNameDuplicatedError
             else:
                 unique_values.add(value)
                 # FIXME can also add reverse link
```
tbittar: The comment may be removed

src/python/antares_xpansion/input_checker.py
```diff
@@ -229,11 +257,25 @@ def check_candidates_file(driver):
...
 
+
+def check_candidate_name_and_link(ini_file):
+    # check that name is not empty and does not have space
```
tbittar, line 237: The comments may be removed

src/python/antares_xpansion/input_checker.py
```diff
@@ -243,13 +285,27 @@ def check_candidates_file(driver):
             if max_units != 0 or unit_size != 0:
                 flushed_print("Illegal values in section %s: cannot assign non-null values simultaneously \
```
tbittar: Why is there a `\` in the message ? This leads to a wide blank when printing the message on my console

src/python/antares_xpansion/input_checker.py
```diff
@@ -243,13 +285,27 @@ def check_candidates_file(driver):
...
                       to max-investment and (unit-size or max_units)" % (each_section))
-                sys.exit(1)
+                raise MaxUnitsAndMaxInvestmentNonNullSimultaneously
         elif max_units == 0 or unit_size == 0:
             flushed_print("Illegal values in section %s: need to assign non-null values to max-investment \
```
tbittar: Same remark as the previous one

src/python/antares_xpansion/input_checker.py
```diff
@@ -260,27 +316,106 @@ def check_candidates_file(driver):
...
+
+def check_candidates_file(candidates_ini_filepath, capacity_dir_path):
+    """
+        checks that a candidate file related to an XpansionDriver has the correct format
+
+        :param driver: the XpansionDriver pointing to the candidates file
+
+        :return: Exits if the candidates files has the wrong format.
+    """
+    default_values = {'name': 'NA',
```
tbittar, line 311: I find it weird that the deprecated options 
   - `enable`
   - `candidate-type`
   - `investment-type`
   - `relaxed`
   - `has-link-profile`
  
 appear here. Are they necessary for the `ConfigParser` initialization and can they be removed ?

src/python/antares_xpansion/input_checker.py
```diff
@@ -260,27 +316,106 @@ def check_candidates_file(driver):
...
+                      'link-profile': '1',
+                      'already-installed-capacity': '0',
+                      'already-installed-link-profile': '1'}
+    ini_file = configparser.ConfigParser(default_values)
+    ini_file.read(candidates_ini_filepath)
+
+    # check attributes types and values
+    check_candidate_attributes(ini_file)
+
+    # check that name is not empty and does not have space
```
tbittar: These comments already appear in the function implementation, they should be removed

src/python/antares_xpansion/input_checker.py
```diff
@@ -260,27 +316,106 @@ def check_candidates_file(driver):
...
+class NotHandledValue(Exception):
+    pass
+
+
+type_str = str
+type_int = int
+type_float = float
+
+# "option": (type, legal_value(s))
+options_types_and_legal_values = {
```
tbittar, line 348: There are also some deprecated options here :
- `method`
- `cut_type`
- `week_selection`
- `relaxed_optimality_gap`

For `master` the option `full_integer` is not valid any more.

src/python/antares_xpansion/input_checker.py
```diff
@@ -292,50 +427,62 @@ def check_setting_option_type(option, value):
...
-    option_type = options_types.get(option)
-    if option_type is None:
+    if options_types_and_legal_values.get(option) is None:
         flushed_print(
             'check_setting_option_type: Illegal %s option in settings file.' % option)
-        sys.exit(1)
-    else:
-        if option_type == 'string':
+        raise NotHandledOption
+
```
tbittar, line 380: Introduce a warning for deprecated options similarly as what is done for the check of candidates (or even raise an exception, easier to do as we just need remove deprecated options from `option_types_and_legal_values`) -> need to be discussed with Mathilde

src/python/antares_xpansion/input_checker.py
tbittar: Allowed float values are in fact positive floats -> find a way to uniformize with the candidates positive float check

src/python/antares_xpansion/input_checker.py
```diff
@@ -292,50 +427,62 @@ def check_setting_option_type(option, value):
...
-                return False
-        elif option_type == 'integer':
-            if value in ["+Inf", "-Inf", "+infini", "-infini"]:
-                return True
+        except ValueError:
+            flushed_print(
+                'check_setting_option_type: Illegal %s option in type, numerical value is expected .' % option)
+            return False
+    elif option_type == type_int:
+        if value in ["+Inf", "-Inf", "+infini", "-infini"]:
```
tbittar: We do not want to do this for  `log_level` and we should not allow `-Inf` or `-infini` for the other parameters

src/python/antares_xpansion/input_checker.py
```diff
@@ -348,82 +495,71 @@ def check_setting_option_value(option, value):
...
-    }
-    legal_values = options_legal_values.get(option)
+    if not check_setting_option_type(option, value):
+        flushed_print(
+            "check_settings : value %s for option %s has the wrong type!" % (value, option))
+        raise OptionTypeError
+
+    legal_values = options_types_and_legal_values.get(option)[1]
 
     skip_verif = ["yearly-weights", "additional-constraints", "solver"]
```
tbittar, line 462: Add deprecated options in `skip_verif`, there is only `master` and `uc_type` left to check

src/python/antares_xpansion/input_checker.py
```diff
@@ -348,82 +495,71 @@ def check_setting_option_value(option, value):
...
     if (option == "optimality_gap") or (option == "relative_gap"):
-        try:
-            gap = float(value)
-            if float(value) >= 0:
-                return True
-            else:
-                raise ValueError
-        except ValueError:
-            print(
+        if float(value) >= 0:
```
tbittar, line 468: Can be done within `check_option_type` if we don as the check for the candidates (the type is positive float)

src/python/antares_xpansion/input_checker.py
```diff
@@ -348,82 +495,71 @@ def check_setting_option_value(option, value):
...
-            except ValueError:
+            max_iter = int(value)
+            if (max_iter == -1) or (max_iter > 0):
+                return True
+            else:
                 flushed_print('Illegal value %s for option %s : only -1 or positive values are allowed'
                               % (value, option))
-                sys.exit(1)
+                raise MaxIterValueError
     elif option == "relaxed_optimality_gap":
```
tbittar: Remove this check of a deprecated option

src/python/antares_xpansion/input_checker.py
```diff
@@ -442,14 +578,9 @@ def check_options(options):
...
-                "check_settings : value %s for option %s has the wrong type!" % (value, option))
-            sys.exit(1)
         check_setting_option_value(option, value)
 
-    if options.get('yearly-weights', "") != "":
-        if options.get("cut_type") == "average":
-            flushed_print(
-                "check_settings : yearly-weights option can not be used when cut_type is average")
-            sys.exit(1)
+    if options.get('yearly-weights', "") != "" and options.get("cut_type") == "average":
```
tbittar: Remove this as `cut_type` is no longer used

tbittar (commented):
The main work to do is to uniformize the way the checks of `candidates.ini` and `settings.ini` are done.

src/python/antares_xpansion/input_checker.py
sgatto: done

src/python/antares_xpansion/input_checker.py
sgatto: I agree. Modification done

src/python/antares_xpansion/input_checker.py
sgatto: corrected

src/python/antares_xpansion/input_checker.py
sgatto: comment removed

src/python/antares_xpansion/input_checker.py
sgatto: checking for loglevel is not problematic

src/python/antares_xpansion/input_checker.py
sgatto: what are the deprecated option ?

src/python/antares_xpansion/input_checker.py
sgatto: removed

sgatto (approved):

src/python/antares_xpansion/input_checker.py
sgatto: removed

src/python/antares_xpansion/input_checker.py
a-zakir: Yes they are checked to inform user to not use them.

PR #357, opened 2022-01-25
allow to print solver's log in file

a-zakir:


src/python/antares_xpansion/benders_driver.py
```diff
@@ -63,6 +70,27 @@ def set_simulation_output_path(self, simulation_output_path: Path):
...
             )
 
+    def _run_and_print_in_std_out_err_and_file(self):
+        with open(self._benders_log_file, "w") as benders_log:
+            with subprocess.Popen(
+                self._get_solver_cmd(), shell=False, stdout=subprocess.PIPE, stderr=subprocess.PIPE,
+            encoding='utf-8') as proc:
+                for output in proc.stdout:
+                    benders_log.write(output)
+
```
sgatto: there should also  be

`sys.stdout.write(output)`

sgatto (approved):

tests/python/test_benders_driver.py
```diff
@@ -208,6 +209,108 @@ def test_clean_benders_step_if_not_keep_mps(self, tmp_path):
...
+        my_install_dir = Path("Dummy/Path/to/")
+        my_n_mpi = 13
+        exe_path = os.path.normpath(os.path.join(my_install_dir, my_benders_mpi))
+
+        benders_driver = BendersDriver(exe_path, "", "")
+
+        simulation_output_path = tmp_path
+        lp_path = Path(os.path.normpath(os.path.join(simulation_output_path, 'lp')))
+        lp_path.mkdir()
+        os.chdir(lp_path)
```
tbittar: Could we refactor this part of code which is the same in the three tests (in a setup method for example) ?

tests/python/test_benders_driver.py
```diff
@@ -208,6 +209,108 @@ def test_clean_benders_step_if_not_keep_mps(self, tmp_path):
...
+                
+
+
+        assert benders_log_file.exists()
+        assert benders_log_file.stat().st_size == len(expected_std_out[0])
+        with open(benders_log_file) as log:
+            lines = [line for line in log.readlines()]
+            assert lines == expected_std_out  
+
+    def test_benders_that_stderr_are_printed_in_log_file(self, tmp_path):
```
tbittar: This test is identical to the previous except that it involves `stderr` in place of `stdout`, is it possible to write a unique test that "takes the stream as parameter" ?

src/python/antares_xpansion/config_loader.py
```diff
@@ -356,6 +356,10 @@ def log_level(self):
             "log_level", self._config.settings_default["log_level"]
         )
         return int(log_level_str) 
+    
+    def benders_log_file(self)-> Path:
```
tbittar, line 360: This function has the same name as `benders_log_file` in `benders_driver.py` line 170, I don't know if this is annoying

PR #358, opened 2022-01-26
add userguide to zip archive

### LINKED ISSUE #320: Add pdf documentation file in the package
Should we put the pdf documentation file in the package ?

sgatto:
This feature modifies the actions in order to add "by hand" the pdf userguide to the zip intaller.

This test aims to close #320

### DIFF
--- a/.github/workflows/build-userguide.yml
+++ b/.github/workflows/build-userguide.yml
@@ -6,6 +6,7 @@ on:
       - main
       - develop
       - doc
+      - test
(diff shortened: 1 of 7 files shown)


sgatto:
I like it very much.
I think that we can now use the same logic to include the userguide in the zip for all the 3 actions used for the releases.

sgatto:
I tested the behavoiur creating a test release (the release has been deleted since)

PR #374, opened 2022-02-16
Sensitivity analysis of investments

tbittar:
Implementation of a sensitivity analysis library, that allows to find investments in the neighbourhood of the optimal solution found by the Benders algorithm


src/cpp/exe/sensitivity/main.cpp
```diff
@@ -0,0 +1,54 @@
...
+            std::cout << desc << std::endl;
+            return 0;
+        }
+
+        po::notify(opts);
+
+        auto sensitivity_input_reader = SensitivityInputReader(json_input_path, benders_output_path, last_master_path, structure_path);
+        SensitivityInputData input_data = sensitivity_input_reader.get_input_data();
+
+        auto writer = std::make_shared<SensitivityWriter>(json_output_path);
```
sgatto: Try is very long, its better to have few lines (the best is only one) in a catch statement.
I propose to split this try statement (remove the variables definitions from a try) inot smaller ones for specific tasks

src/cpp/multisolver_interface/SolverCplex.cpp
```diff
@@ -298,6 +298,12 @@ void SolverCplex::chg_obj(const std::vector<int>& mindex, const std::vector<doub
     zero_status_check(status, "change obj");
 }
 
+void SolverCplex::chg_obj_sense(const bool minimize) {
+	int objsense = minimize ? CPX_MIN : CPX_MAX;
```
sgatto: sense? di you mean direction or sign plus2minus ?
sense seems like meaning :)

PR #375, opened 2022-02-21
Feature/solver log in file

### LINKED ISSUE #341: Write solver logs into a file
Is it possible to write the solver logs, generated when `log_level > 0` into a file ?

a-zakir:


src/cpp/benders/benders_core/include/core/ILogger.h
```diff
@@ -55,6 +55,8 @@ struct LogData {
   double max_iterations;
 };
 
+const std::string indent_0 = "\t\t";
+const std::string indent_1 = "\t";
```
sgatto: I think that these constants should be where the logger are implemented. These are implementation details that belongs to the classes that actually print

sgatto (approved):

PR #383, opened 2022-02-25
unify actions release and standard CI

sgatto:
reduce redundunce in the code of the github actions


.github/workflows/centos7-system.yml
```diff
@@ -83,65 +121,102 @@ jobs:
...
-           cd _build
-           cpack3 -G RPM
-           
-    - name: Installer .tar.gz creation
-      run: |
-           cd _build
-           cpack3 -G TGZ
 
-    - name: Installer archive upload
+    - name: Umpload  .tar.gz
```
JasonMarechal25: I think there's a typo in the name, you probably meant "Upload"

sgatto: you're right. Its been corrected

PR #412, opened 2022-03-25
Fix/no solver specified in settings file

### LINKED ISSUE #410: Xpansion fails when no solver is specified in `settings.ini`
When no solver is given in `settings.ini`, we expect xpansion to use the default `Cbc` solver. However, even though a message states that `Cbc` is used, Xpansion fails later on with the message `Solver ' ' not available`.

The problem probably comes from the function https://github.com/AntaresSimulatorTeam/antares-xpansion/blob/71cfdea92810c698ecc3d94bc48ca561334d0e6c/src/python/antares_xpansion/xpansion_study_reader.py#L29-L40 that should replace an empty solver string with `Cbc` similarly as what is done in https://github.com/AntaresSimulatorTeam/antares-xpansion/blob/71cfdea92810c698ecc3d94bc48ca561334d0e6c/src/python/antares_xpansion/xpansion_study_reader.py#L73-L86

a-zakir:


src/python/antares_xpansion/xpansion_study_reader.py
```diff
@@ -78,20 +83,22 @@ def check_solver(solver_str: str, available_solvers: List[str]):
...
-            print("No solver defined in user/expansion/settings.ini. Cbc used")
-            solver_str = "Cbc"
+            raise XpansionStudyReader.NoSolverValue(
+                f"Error in solver definition, please check user/expansion/settings.ini file")
         if not available_solvers.count(solver_str):
             raise XpansionStudyReader.SolverNotAvailable(
                 f'Solver {solver_str} not available. Please use one of these solver in user/expansion/settings.ini : {available_solvers}')
 
-    @staticmethod
+    @ staticmethod
```
JasonMarechal25: Is the extra space wanted ? Won't it cause an issue ?

a-zakir: im surprised as you! it seems to be accepted, i think my auto-formatter did this.

PR #421, opened 2022-03-29
Fix/windows ci

JasonMarechal25:


.gitmodules
```diff
@@ -1,7 +1,6 @@
 [submodule "antares-deps"]
 	path = antares-deps
 	url = https://github.com/AntaresSimulatorTeam/antares-deps.git
-	branch = v2.0.1
```
a-zakir: why delete this line? without setting the commit id or the branch, the submodule will point on what?

a-zakir (approved):

PR #429, opened 2022-03-31
Fix action for dependency compilation

JasonMarechal25:


.github/workflows/ubuntu-system-deps-build.yml
```diff
@@ -6,6 +6,7 @@ on:
       - main
       - develop
       - release/*
+      - *
```
a-zakir: this action will be run for every branch?

JasonMarechal25: Yes. I should have put a comment. It's to trigger the action for this branch and check that the action does pass.
Since it's green I will now remove this commit

a-zakir (approved):

PR #441, opened 2022-04-19
Write optimal basis after Benders and use it for sensitivity

### LINKED ISSUE #440: Save optimal basis after Benders resolution
In order to speed up sensitivity analysis, it can be profitable to provide an initial (feasible) basis. We can easily get one by saving the optimal basis of the last master solved in the Benders step and loading it before launching the sensitivity study.

As the sensitivity problems only differ from the last master by a modification in the objective coefficients and the addition of a single constraint, it is easy to construct a initial basis for the sensitivity study from the one saved after Benders.

tbittar:


JasonMarechal25 (approved):

src/cpp/benders/benders_core/BendersBase.cpp
```diff
@@ -699,3 +699,10 @@ double BendersBase::GetSubproblemCost() const { return _data.subproblem_cost; }
 void BendersBase::SetSubproblemCost(const double &subproblem_cost) {
   _data.subproblem_cost = subproblem_cost;
 }
+
+void BendersBase::write_basis() const {
+  std::filesystem::path const filename(
```
a-zakir: std::filesystem::path is redundant, we may use `const auto filename = ... `

src/cpp/multisolver_interface/SolverCplex.h
```diff
@@ -74,9 +74,12 @@ class SolverCplex : public SolverAbstract {
  public:
   virtual void write_prob_mps(const std::filesystem::path &filename) override;
   virtual void write_prob_lp(const std::filesystem::path &filename) override;
+  virtual void write_basis(const std::filesystem::path &filename) override{};
 
   virtual void read_prob_mps(const std::filesystem::path &filename) override;
-  virtual void read_prob_lp(const std::string &filename) override;
+  virtual void read_prob_lp(const std::filesystem::path &filename) override;
```
a-zakir, line 80: i think `virtual` keyword is not needed here.

a-zakir (changes requested):

src/cpp/multisolver_interface/SolverCplex.h
tbittar: Actually, all `virtual` keywords in files from the child classes of `SolverAbstract` are not needed. I keep it here to be coherent with the rest of the file, but I can remove it everywhere if you think it is worth it.

a-zakir: Your are right and the pr is already large, let's do that in dedicated branch.

a-zakir (approved):

PR #449, opened 2022-04-29
Handle different chronicle depending on MC year

### LINKED ISSUE #205: Link-profile differentiated by Monte-Carlo years
Currently, the link-profile file (located in the _study/user/expansion/capa_ folder) contains one or two columns to provide direct and indirect links capacities. To represent different derating, temporary lower availability or load factor, the profile is deterministic: the same profile will be used by Xpansion for all Monte-Carlo years of the Antares study.

image

The idea would be to have a different link-profile per Monte-Carlo year .

Like for **_yearly-weights_**, we assume that the user has put either 1 column valid for all Monte-Carlo years, or the right number of columns compared to the number of years of the study.

Should we keep only one file? It seems ambiguous to keep only one file because we cannot distinguish the direct+indirect case and the direct case with several years, this change will have an impact on the existing studies...

For example, we could have :

- 1 file for the direct capacities, containing either 1 single column valid for all the years, or 1 column per year and it is up to the user to put the right number of years. In this case, a test should be done to check if there is only one or several columns and if there are several columns, check that it corresponds to the number of years of the study.
- Similarly, 1 file for the indirect abilities.

The _**lp_namer**_ code should be modified : 
- p[t] is the power of a link at each time step.
- pmax is the capacity of the candidate for a link we are testing
- alpha[t,**year**] is the percentage corresponding to the _link_profile_
- alpha0[t,**year**] is the percentage corresponding to the _already-installed-link-profile_
- pMax0 is the _already-installed-capacity_
- the constraint is :  `p[t] - alpha[t,**year**].pMax - alpha0[t,**year**].pMax0 <= 0`

Updating the study at the end of Xpansion would not be possible because link capacities cannot be differentiated by Monte Carlo year in Antares. Only an average capacity coulb be set.

JasonMarechal25:


src/cpp/lpnamer/input_reader/LinkProfileReader.cpp
```diff
@@ -2,54 +2,91 @@
 
 #include <filesystem>
 
-LinkProfile LinkProfileReader::ReadLinkProfile(
-    const std::filesystem::path &filename) {
+std::vector<LinkProfile> LinkProfileReader::ReadLinkProfile(
+    const std::filesystem::path &direct_filename, const std::filesystem::path& indirect_file_name) {
+  if (std::ifstream infile(direct_filename); !infile.good()) {
+    throw std::filesystem::filesystem_error("unable to open file",
```
a-zakir: direct-filename is tested ok, and what for indirect-filename?

a-zakir: direct-filename is tested ok, and what for indirect-filename?

src/cpp/lpnamer/model/ActiveLinks.cpp
```diff
@@ -162,11 +186,27 @@ const std::vector<Candidate>& ActiveLink::getCandidates() const {
...
 double ActiveLink::already_installed_indirect_profile(size_t timeStep) const {
-  return _already_installed_profile.getIndirectProfile(timeStep);
+  return _already_installed_profile.at(0).getIndirectProfile(timeStep);
+}
+
+double ActiveLink::already_installed_direct_profile(size_t year, size_t timeStep) const {
+  if (year == 0)
+    year = 1;
+  if (year > _already_installed_profile.size())
+    year = 1;
```
a-zakir: is the case `year> _already_installed_profile.size()` possible?
this function is only used in test lib

a-zakir: is the case `year> _already_installed_profile.size()` possible?
this function is only used in test lib

src/cpp/lpnamer/model/Candidate.cpp
```diff
@@ -1,19 +1,37 @@
...
+double Candidate::indirectCapacityFactor(size_t timeStep) const {
+  return _profile.at(0).getIndirectProfile(timeStep);
+}
+
+double Candidate::directCapacityFactor(size_t chronicle_number, size_t timeStep) const {
+  if (chronicle_number == 0) {
+    return _profile.at(0).getDirectProfile(timeStep);
+  } else {
+    //1-based chronicle in 0 based vector
+    return _profile.at(chronicle_number - 1).getDirectProfile(timeStep);
```
a-zakir: in case of multiple chronicles `chronicle_number=0` and `chronicle_number=1` gives same result, normal?

a-zakir: in case of multiple chronicles `chronicle_number=0` and `chronicle_number=1` gives same result, normal?

src/cpp/lpnamer/problem_modifier/CMakeLists.txt
```diff
@@ -21,7 +21,9 @@ add_library (lp_namer_problem_modifier STATIC
 	${CMAKE_CURRENT_SOURCE_DIR}/LauncherHelpers.h
 	${CMAKE_CURRENT_SOURCE_DIR}/MasterGeneration.cpp
 	${CMAKE_CURRENT_SOURCE_DIR}/MasterGeneration.h
-    )
+        ../model/Problem.cpp ../model/Problem.h
```
a-zakir: theses src are added twice, see in /model.

May be consider adding them in /model/CMakeLists.txt  with relative path: 
`${CMAKE_CURRENT_SOURCE_DIR}Problem.h`
`${CMAKE_CURRENT_SOURCE_DIR}Problem.cpp`
then they will be visible in /problem_modifier lib

a-zakir: theses src are added twice, see in /model.

May be consider adding them in /model/CMakeLists.txt  with relative path: 
`${CMAKE_CURRENT_SOURCE_DIR}Problem.h`
`${CMAKE_CURRENT_SOURCE_DIR}Problem.cpp`
then they will be visible in /problem_modifier lib

tests/python/test_chronicles_checker.py
```diff
@@ -0,0 +1,48 @@
...
+        checker = ChronicleChecker(study_path)
+        with pytest.raises(expected_exception) as exc:
+            checker.CheckChronicleConstraints()
+
+    def test_all_ok_no_installed_profile(self, tmp_path):
+        study_path = DATA_PATH / "all_ntc_ok_no_installed_profile"
+        checker = ChronicleChecker(study_path)
+        try:
+            checker.CheckChronicleConstraints()
+        except Exception as exc:
```
a-zakir, line 52: is the exception is unknown? if no it's we may use pytest.raises to catch it

src/cpp/lpnamer/model/Candidate.cpp
JasonMarechal25: Added justification in code to explain. Given the length of the justification it bring to light a potential design flaw

  /* When there is no scenario builder output from antares. The chronicle map contains only zeros
   * When there is an output the map contains values ranging from 1 to <number of chronicles>
   * The first case happen when there is only one profile, so we need to return it
   * The second case happen when you have multiple profiles, in this case we need to 
   * project values from [1,N] to [0,N-1] to return the proper profiles
   */

src/cpp/lpnamer/problem_modifier/CMakeLists.txt
JasonMarechal25: Thank you, it's indeed better

tests/python/test_chronicles_checker.py
JasonMarechal25: Several exception types have been defined. In this test we are not testing that an exception occur but that no exception occurs, be it the one defines for the SUT or even generic ones like IOAccess

src/cpp/lpnamer/model/ActiveLinks.cpp
JasonMarechal25: Yes it happens in cases where no profile is defined.
For example a user can define a link profile but no installed profile or vice versa. When nothing is defined a chronicle filled with ones is generated by default.
Maybe a proper solution would be to have the proper number of chronicles generated given the parameters of the study.

docs/user-guide/get-started/candidate-definition.md
```diff
@@ -209,6 +216,21 @@ obtained by Antares-Xpansion in order to obtain the real production
 program with outages and RES intermittence varying according to the
 scenarios.
 
+## Using different profile depending on Monte-Carlo year
+With Antares simulator 8.2 users can define different NTC chronicles for different Monte-Carlo year. Briefly, users defines any number
```
flomnes:
```suggestion
From Antares_Simulator 8.2, users can define different NTC chronicles for different Monte-Carlo years. Briefly, users defines any number
```

docs/user-guide/get-started/candidate-definition.md
```diff
@@ -209,6 +216,21 @@ obtained by Antares-Xpansion in order to obtain the real production
...
+of chronicles for each link. For each link users can select or define randomness to select one 
+of the said chronicles in a result we call "scenario"
+(more information in antares simulator documentation). In Antares Xpansion it is possible to use 
+those scenarios to select different link profiles for a given Monte-Carlo year. There are a few pre-requisite:
+1) On a candidate, direct and indirect link profiles must have the same number of column
```
flomnes:
```suggestion
1) On a candidate, direct and indirect link profiles must have the same number of columns
```

docs/user-guide/get-started/candidate-definition.md
```diff
@@ -209,6 +216,21 @@ obtained by Antares-Xpansion in order to obtain the real production
...
+2) The same rule applies for already installed direct and indirect profile
```
flomnes:
```suggestion
2) The same rule applies for already installed direct and indirect profiles
```

docs/user-guide/get-started/candidate-definition.md
```diff
@@ -209,6 +216,21 @@ obtained by Antares-Xpansion in order to obtain the real production
...
+3) The number of column in link profile and already installed profiles must be the same
+4) The number of column for NTC chronicles on a given link must be the same as the number of column
```
flomnes:
```suggestion
3) The number of columns in link profile and already installed profiles must be the same
4) The number of columns for NTC chronicles on a given link must be the same as the number of columns
```

src/cpp/helpers/AntaresVersionProvider.cpp
```diff
@@ -0,0 +1,33 @@
...
+  std::ifstream studyFile_l(study_path / STUDY_FILE);
+  std::string line_l;
+  const std::string versionProperty_l("version = ");
+
+  int version = DEFAULT_ANTARES_VERSION;
+  while (std::getline(studyFile_l, line_l)) {
+    size_t pos_l = line_l.find(versionProperty_l);
+    if (pos_l != std::string::npos) {
+      std::stringstream buffer(line_l.substr(pos_l + versionProperty_l.size()));
+      buffer >> version;
```
flomnes, line 25: This conversion could fail, right ? For example if **study.antares** contains 
```
version = garbage
```

flomnes: It seems that in such case, `version` is set to 0 and failbit is set to 1. Maybe we should handle that case.

[source](https://en.cppreference.com/w/cpp/io/basic_istream/operator_gtgt)

flomnes, line 25:
```suggestion
      buffer >> version;
      if (buffer.fail()) {
          version = DEFAULT_ANTARES_VERSION;
      }
```

src/cpp/lpnamer/input_reader/LinkProfileReader.cpp
```diff
@@ -2,54 +2,96 @@
...
-      throw std::runtime_error("error not enough line in link-profile " +
-                               filename.string());
+      throw std::domain_error("error not enough line in link-profile " +
+                              filename.string());
     }
   }
   infile.close();
+}
+void LinkProfileReader::UpdateProfile(std::vector<LinkProfile> &result,
+                                      bool directProfile, double value,
```
flomnes, line 58: Use an enum-class instead of a bool
```cpp
enum class LinkDirection {
   Direct,
   Indirect
};
```

flomnes: `line_id` doesn't make sense in this context. I think that this index refers to an hour, and should be named as such. The other parameter `chroniche_id` has a better name for example.

src/cpp/lpnamer/input_reader/LinkProfileReader.cpp
```diff
@@ -2,54 +2,96 @@
...
+    profile.indirect_link_profile.at(line_id) = value;
+  }
+}
 
-  return result;
+void LinkProfileReader::UpdateResultWithMissingChronicle(
+    std::vector<LinkProfile> &result, int chronicle_id) {
+  if (result.size() <= chronicle_id) {
+    result.emplace_back();
+  }
```
flomnes: It seems that we only handle one missing column. What if more of them are missing ?

src/cpp/lpnamer/input_reader/LinkProfileReader.cpp
```diff
@@ -2,54 +2,96 @@
...
+                  candidate_data.installed_direct_link_profile_name, candidate_data.installed_indirect_link_profile_name);
+    importProfile(mapLinkProfile, capacity_folder,
+                  candidate_data.direct_link_profile, candidate_data.indirect_link_profile);
   }
   return mapLinkProfile;
 }
 
 void LinkProfileReader::importProfile(
-    std::map<std::string, LinkProfile> &mapLinkProfile,
+    std::map<std::string, std::vector<LinkProfile>> &mapLinkProfile,
```
flomnes, line 89: Define custom types ?

src/cpp/lpnamer/model/ActiveLinks.cpp
```diff
@@ -11,22 +12,33 @@ bool doubles_are_different(const double a, const double b) {
...
-      candidate_data, getProfileFromProfileMap(candidate_data.link_profile));
+      candidate_data, getProfilesFromProfileMap(candidate_data.direct_link_profile));
 }
 
 ActiveLinksBuilder::ActiveLinksBuilder(
-    const std::vector<CandidateData>& candidateList,
-    const std::map<std::string, LinkProfile>& profile_map)
-    : _candidateDatas(candidateList), _profile_map(profile_map) {
+    std::vector<CandidateData>  candidateList,
+    std::map<std::string, std::vector<LinkProfile>>  profile_map,
```
flomnes, line 20: I understand that you're using `std::move` to grab ownership to the object and avoid copies. However, the copy constructor is called when passing arguments by value. I suggest passing value by non-const reference instead, so that only the move constructor is called.

Pass args by value :
1. The copy constructor is called
2. The move constructor is called

Pass args by non-const reference
1. The move constructor is called

```cpp
#include <iostream>

struct Example {
  Example() = default;
  Example(Example& ex) {
    std::cout << "Example(Example& ex) COPY NOT GOOD\n";
  }
  Example(const Example& ex) {
    std::cout << "Example(const Example& ex)\n";
  }
  Example(Example&& ex) {
    std::cout << "Example(Example&& ex)\n";
  }
};

struct ExampleGrabByValue {
  ExampleGrabByValue(Example ex) : ex(std::move(ex)) {}
  Example ex;
};

struct ExampleGrabByRef {
  ExampleGrabByRef(Example& ex) : ex(std::move(ex)) {}
  Example ex;
};

int main() {
  Example ex;
  std::cout << "ExampleGrabByValue\n";
  ExampleGrabByValue byValue(ex);
  std::cout << "ExampleGrabByRef\n";
  ExampleGrabByRef byRef(ex);

  // Possible output
  // ExampleGrabByValue
  // Example(Example& ex) COPY NOT GOOD
  // Example(Example&& ex)
  // ExampleGrabByRef
  // Example(Example&& ex)
}
```

src/cpp/lpnamer/model/ActiveLinks.cpp
```diff
@@ -116,42 +128,54 @@ int ActiveLinksBuilder::getLinkIndexOf(int link_id) const {
...
+  if (profiles.empty())
+    return {LinkProfile()};
+  return profiles;
+}
+
+ActiveLink::ActiveLink(
+    int idLink, std::string  linkName, std::string  linkor,
+    std::string  linkex, const double& already_installed_capacity,
+    std::map<unsigned int, unsigned int> mc_year_to_chronicle)
+    : mc_year_to_chronicle_(std::move(mc_year_to_chronicle)),
```
flomnes, line 157: Same remark as above : pass arguments by non-const reference.

src/cpp/lpnamer/model/ActiveLinks.cpp
```diff
@@ -162,11 +186,25 @@ const std::vector<Candidate>& ActiveLink::getCandidates() const {
...
+    chronicle_number = 1;
+  return _already_installed_profile.at(chronicle_number - 1).getDirectProfile(timeStep);
+}
+
+double ActiveLink::already_installed_indirect_profile(size_t chronicle_number, size_t timeStep) const {
+  if (chronicle_number == 0) chronicle_number = 1;
+  if (chronicle_number > _already_installed_profile.size())
+    chronicle_number = 1;
+  return _already_installed_profile.at(chronicle_number - 1).getIndirectProfile(timeStep);
 }
```
flomnes, line 208: To avoid code duplication, I suggest using a single function for both directions, passing enum-class LinkDirection (direct/indirect)

src/cpp/lpnamer/model/Candidate.cpp
```diff
@@ -1,19 +1,49 @@
 #include "Candidate.h"
 
-Candidate::Candidate(const CandidateData& data, const LinkProfile& profile)
-    : _profile(profile),
+#include <utility>
+
+Candidate::Candidate(const CandidateData& data, std::vector<LinkProfile>  profile)
+    : _profile(std::move(profile)),
```
flomnes, line 7: pass by ref

src/cpp/lpnamer/model/Candidate.cpp
```diff
@@ -1,19 +1,49 @@
...
 
-double Candidate::direct_profile(size_t timeStep) const {
-  return _profile.getDirectProfile(timeStep);
+double Candidate::directCapacityFactor(size_t timeStep) const {
+  return _profile.at(0).getDirectProfile(timeStep);
+}
+
+double Candidate::indirectCapacityFactor(size_t timeStep) const {
+  return _profile.at(0).getIndirectProfile(timeStep);
+}
```
flomnes: Same as above : one single function, an additional parameter for the direction.

src/cpp/lpnamer/model/Candidate.cpp
```diff
@@ -35,4 +65,14 @@ bool Candidate::is_integer() const { return _unit_size != 0.0; }
...
+void Candidate::set_name(const std::string& name) { _name = name; }
+
+bool Candidate::hasNullProfile(unsigned int chronicle,
+                               const std::set<int>& time_steps) const {
+  bool hasOnlyNullProfile = true;
+  for (int time_step: time_steps) {
+    hasOnlyNullProfile &= directCapacityFactor(chronicle, time_step) == 0.0;
+    hasOnlyNullProfile &= indirectCapacityFactor(chronicle, time_step) == 0.0;
+  }
+  return hasOnlyNullProfile;
```
flomnes: Use [`std::all_of`](https://en.cppreference.com/w/cpp/algorithm/all_any_none_of)

src/cpp/lpnamer/model/Candidate.h
```diff
@@ -30,10 +32,13 @@ struct CandidateData {
 class Candidate {
  public:
   Candidate() = default;
-  Candidate(const CandidateData& data, const LinkProfile& profile);
+  Candidate(const CandidateData& data, std::vector<LinkProfile>  profile);
```
flomnes, line 35: pass by ref

src/cpp/lpnamer/model/ChronicleMapReader.h
```diff
@@ -0,0 +1,32 @@
...
+                                    unsigned int montecarlo_year) const;
+  void AssignChronicleValuesToMCYears(std::stringstream& ss) const;
+
+ public:
+  [[nodiscard]] std::map<unsigned int, unsigned int> read(std::string const& input) const;
+
+  [[nodiscard]] std::map<unsigned int, unsigned int> read(std::stringstream& input) const;
+
+  [[nodiscard]] std::map<unsigned int, unsigned int> read(
+      const std::ifstream& input) const;
```
flomnes: You have 3 versions of `read`. I suggest that you remove the `std::stringstream&` and the `const std::ifstream&` versions, and create a generic `std::istream&` instead.

Example
```cpp
#include <iostream>
#include <sstream>
#include <fstream>
#include <cassert>

int read(std::istream& is) {
  int out;
  is >> out;
  return out;
}

int main() {
  {
    std::string s("1");
    std::stringstream ss(s);
    auto a = read(ss);
    assert(a == 1);
  }

  {
    std::ifstream f("test.txt"); // 2
    auto a = read(f);
    assert(a == 2);
  }
}
```

flomnes (commented):
I haven't read everything, about 1/2 of the cpp code only. I'll continue later

src/cpp/lpnamer/model/Candidate.cpp
JasonMarechal25: Fixed, thanks

src/cpp/helpers/AntaresVersionProvider.cpp
JasonMarechal25: I'm assuming that simulator always generate a proper output that's why I don' check for it.

src/cpp/lpnamer/input_reader/LinkProfileReader.cpp
JasonMarechal25: Done, thank you

src/cpp/lpnamer/input_reader/LinkProfileReader.cpp
JasonMarechal25: Renammed line_id as time_step to be more expressive

src/cpp/lpnamer/model/ActiveLinks.cpp
JasonMarechal25: It is "intented".
With the existing design the different values _candidateList_ or _profile_map_ are copied in class members. We need this copy (or need to make more profound changes).
The change was mainly prompted by sonar lint but it add expressiveness to the interface.

`void foo(const Bar& bar)` means that I take a read only reference to a existing value and hide the fact that we will copy it for ourself.
`void foo(Bar bar)`means that I need a value and express that if it already exist will at least be copied on the stack.

src/cpp/lpnamer/model/ActiveLinks.cpp
JasonMarechal25: It is a good suggestion. I'd prefer not to implement it for now because the whole profile management system (or even lp namer) need a good rethinking. We often work on both values (direct and indirect) at a time so it would be good to think of an interface that make it seamless for clients if it is possible.

src/cpp/lpnamer/model/ChronicleMapReader.h
JasonMarechal25: Fixed with only 2 version. One with streams, one with string

a-zakir (approved):

src/cpp/helpers/AntaresVersionProvider.cpp
JasonMarechal25: Fix with proposed suggestion

src/cpp/lpnamer/model/Problem.h
```diff
@@ -0,0 +1,108 @@
...
+  private:
+  const SolverAbstract::Ptr solver_abstract_;
+
+ public:
+  [[nodiscard]] unsigned int McYear() const { return mc_year; }
+
+  unsigned int mc_year = 0;
+
+  int get_number_of_instances() override { return solver_abstract_->get_number_of_instances(); }
+  [[nodiscard]] [[nodiscard]] std::string get_solver_name() const override { return solver_abstract_->get_solver_name(); }
```
flomnes:
```suggestion
  [[nodiscard]] std::string get_solver_name() const override { return solver_abstract_->get_solver_name(); }
```

src/cpp/lpnamer/model/Problem.h
```diff
@@ -0,0 +1,108 @@
...
+class Problem: public SolverAbstract
+{
+ public:
+  Problem() = delete;
+  explicit Problem(SolverAbstract::Ptr solver_abstract)
+  : solver_abstract_(std::move(solver_abstract))
+  {
+  }
+  private:
+  const SolverAbstract::Ptr solver_abstract_;
```
flomnes, line 24: I guess that you wanted to add `McYear` to class `SolverAbstract` to change the behavior of method `read_prob_mps` (deduce & store the MC year from the filename). So you 
- Inherit from this class (l.15)
- Keep a pointer to it (l.24)
- Override all methods from this class to call the same methods on this pointer (l.33-105, see also Problem.cpp)

My question is : why do you need to keep this pointer and override all those methods ? Why not something like this
```cpp
class Problem : public SolverAbstract {
public:    
    int mc_year;
    void read_prob_mps(/* type */ filename) override {
    // See Problem.cpp
    }
    // Don't override other methods. Class Problem is still pure abstract.
};
```
You'd only need to replace `SolverAbstract`  with `Problem` where needed.

src/cpp/lpnamer/problem_modifier/CMakeLists.txt
```diff
@@ -21,7 +21,8 @@ add_library (lp_namer_problem_modifier STATIC
 	${CMAKE_CURRENT_SOURCE_DIR}/LauncherHelpers.h
 	${CMAKE_CURRENT_SOURCE_DIR}/MasterGeneration.cpp
 	${CMAKE_CURRENT_SOURCE_DIR}/MasterGeneration.h
-    )
+	${CMAKE_CURRENT_SOURCE_DIR}/LinkParametersCSVOverwriter.h ${CMAKE_CURRENT_SOURCE_DIR}/LinkParametersCSVOverwriter.cpp
+		LinkdataRecord.h LinkCapacitiesCSVWriter.h LinkCapacitiesCSVWriter.cpp StudyUpdateStrategy.cpp StudyUpdateStrategy.h StudyUpdateLinkCapacitiesStrategy.cpp StudyUpdateLinkCapacitiesStrategy.h StudyUpdateLinkParameterStrategy.cpp StudyUpdateLinkParameterStrategy.h)
```
flomnes: Either :
- Remove `${CMAKE_CURRENT_SOURCE_DIR}` prefix from existing files
- Add it to the new files

In any case : 1 file per line.

src/cpp/lpnamer/problem_modifier/LinkParametersCSVOverwriter.cpp
```diff
@@ -0,0 +1,59 @@
...
+              << std::endl;
+    return false;
+  }
+
+  already_warned_ = false;
+  if (!inputCsv_l_.is_open()) {
+    std::cout << "WARNING: Missing file to update antares study : "
+              << linkdataFilename_l << "."
+              << " Unknown valus were populated with 0 in a new created file."
+              << std::endl;
```
flomnes:
```suggestion
    std::cout << "WARNING: Missing file to update antares study : "
              << linkdataFilename_l << "."
              << " Unknown values were populated with 0 in a newly created file."
              << std::endl;
```
Publish in logs ?

src/cpp/lpnamer/problem_modifier/ProblemModifier.cpp
```diff
@@ -25,26 +15,22 @@ std::set<int> extract_time_steps(
...
+    auto candidates = link.getCandidates();
+    auto newEnd = std::remove_if(
+        candidates.begin(), candidates.end(),
+        [&](const Candidate &cand) {
+          return cand.hasNullProfile(chronicleToUse(link), time_steps);
+        });
+
+    std::move(candidates.begin(), newEnd, std::back_inserter(candidates_to_keep));
+  }
+  return candidates_to_keep;
```
flomnes: If my understanding is correct, you

1. Run a loop through active links
2. For each active link, you select candidates that satisfy a `predicate`
3. Concatenate all the candidates, then `return` them

It seems that you create a copy of the candidates at l.24. You could probably do something like
```cpp
std::vector<Candidate> candidates_to_keep;
for (const auto link& : active_links) // you may use std::for_each
   for (const auto candidate& : link.getCandidates()) // I assume that getCandidates returns a const-ref
       if (predicate(candidate)) {
           candidates_to_keep.push_back(candidate);
       }
   }
   return candidates_to_keep;
}
```
Possibly keep `predicate` in a lambda-function. IM your intention would look more clear than using `<algorithm>` dark magic.

src/cpp/lpnamer/problem_modifier/ProblemModifier.cpp
```diff
@@ -206,65 +192,83 @@ void ProblemModifier::add_direct_profile_column_constraint(
...
-      dmatval.push_back(-candidate.direct_profile(column.time_step));
+      dmatval.push_back(-candidate.directCapacityFactor(chronicle_to_use, column.time_step));
     }
   }
   rstart.push_back((int)dmatval.size());
 }
+bool ProblemModifier::candidateContributionIsNotNull(const ColumnToChange &column,
+                                             unsigned int chronicle_to_use,
+                                             const Candidate &candidate) const {
+  return candidate.directCapacityFactor(chronicle_to_use, column.time_step) != 0.0;
```
flomnes, line 213: Why specifically use the direct capacity ? Everything should be symmetric, _ie_ swapping direct <-> indirect should not change the behavior.

src/cpp/lpnamer/problem_modifier/ProblemModifier.cpp
```diff
@@ -206,65 +192,83 @@ void ProblemModifier::add_direct_profile_column_constraint(
...
+      link.already_installed_indirect_profile(chronicle_to_use, column.time_step);
   rhs.push_back(-already_installed_capacity *
                 indirect_already_installed_profile_at_timestep);
 
   rowtype.push_back('G');
   colind.push_back(column.id);
   dmatval.push_back(1);
   for (const auto &candidate : link.getCandidates()) {
-    if (candidate.indirect_profile(column.time_step) != 0.0) {
+    if (candidate.indirectCapacityFactor(chronicle_to_use, column.time_step) != 0.0) {
```
flomnes: You use a dedicated method `candidateContributionIsNotNull` above, why not here ?

src/cpp/lpnamer/problem_modifier/ProblemModifier.cpp
```diff
@@ -206,65 +192,83 @@ void ProblemModifier::add_direct_profile_column_constraint(
...
-      dmatval.push_back(candidate.indirect_profile(column.time_step));
+      dmatval.push_back(candidate.indirectCapacityFactor(chronicle_to_use, column.time_step));
     }
   }
   rstart.push_back((int)dmatval.size());
 }
+unsigned int ProblemModifier::chronicleToUse(const ActiveLink &link) const {
+  auto chronicle_map = link.McYearToChronicle();
+  auto problem_mc_year = _math_problem->McYear();
+  auto chronicle_to_use = chronicle_map[problem_mc_year];
```
flomnes, line 250: Careful, this may return the default value for `V` in `std::map<K, V>` if key is not found (probably 0 here). I'd recommend using `std::map<K,V>::at` if this is not what you want.

src/cpp/lpnamer/problem_modifier/StudyUpdateLinkCapacitiesStrategy.cpp
```diff
@@ -0,0 +1,60 @@
...
+#include "LinkCapacitiesCSVWriter.h"
+
+int StudyUpdateLinkCapacitiesStrategy::UpdateLinkCapacities(
+    const ActiveLink& link_p,
+    const std::map<std::string, double>& investments_p) const {
+  LinkCapacitiesCSVWriter capacities_csv_writer(link_p,
+                                                antares_link_folder_path);
+
+  for (int i = 0; i < NUMBER_OF_HOUR_PER_YEAR; ++i) {
+    if (std::vector<std::pair<double, double>> chronicle_capacities_at_time =
```
flomnes:
```suggestion
    if (auto chronicle_capacities_at_time =
```

src/cpp/lpnamer/problem_modifier/StudyUpdateLinkParameterStrategy.cpp
```diff
@@ -0,0 +1,70 @@
...
+    EnsureCandidateInvestmentFound(investments_p, link_p, candidate,
+                                   it_candidate);
+    double candidate_investment = it_candidate->second;
+
+    direct_l +=
+        candidate_investment * candidate.directCapacityFactor(timepoint_p);
+    indirect_l +=
+        candidate_investment * candidate.indirectCapacityFactor(timepoint_p);
+  }
+  return std::make_pair(direct_l, indirect_l);
```
flomnes: Since C++17 I think that template argument deduction rules allow
```suggestion
  return std::pair(direct_l, indirect_l);
```

If my memory is correct, helper `make_pair` was used because only functions were supported for argument deduction. See [here](https://en.cppreference.com/w/cpp/language/class_template_argument_deduction).

src/cpp/lpnamer/input_reader/LinkProfileReader.cpp
JasonMarechal25: Renamed as "ConstructChronicle"
When reading link profile files we do it in two steps, one for direct values, one for indirect values.
We don't know the number of chronicle in advance.
The first time we read a new column (chronicle_id) we need to construct a default profile that will be updated with the proper values. The first time occur when reading the column of the direct file, the second time for indirect file. The second time we don't need to construct the chronicle (LinkProfile object) just update the indirect values.

JasonMarechal25: _Result.size_ will play catch up with _chronicle_id_. Every step of the loop, chronicle_id is incremented, so for the "direct pass", every next step of the loop _result.size_ is one item fewer than _chronicle_id._

src/cpp/lpnamer/input_reader/LinkProfileReader.cpp
JasonMarechal25: Done, thank you.
Both file are validated with new function _EnsureFileIsGood_

src/cpp/lpnamer/model/Problem.h
JasonMarechal25: Fixed up the commit where it was introduced, thank you.

src/cpp/lpnamer/model/Problem.h
JasonMarechal25: Problem is a decorator.
It decorates a object of the SolverAbstract hierarchy with the mc_year property (and read_prob_mps). That means an instance of Problem can decorate any of SolverCBC, SolverCLP or SolverXpress without altering the behaviour of those class.
It can't be pure virtual since it need to be instantiated.
It does override the methods but more than this, it forwards the behaviour to the SolverAbstract instance it decorates.

src/cpp/lpnamer/problem_modifier/CMakeLists.txt
JasonMarechal25: Fixed with prefix for all and reformat file

src/cpp/lpnamer/problem_modifier/LinkParametersCSVOverwriter.cpp
JasonMarechal25: It would be better yes. Here I don't introduce code, I just moved it. (doesn't mean I didn't add cout elsewhere).
Make a proper usage of a logger averywhere would be a task in itself I think

src/cpp/lpnamer/problem_modifier/ProblemModifier.cpp
JasonMarechal25: "Black magic" was already there. I kept the usage of std::algorithm but I used your algorithm instead of the one in place.
The one in place: copy all candidates, remove those with null profile.
Your algorithm: look at all candidates, copy (push back) if not null profile
=> std::copy_if

src/cpp/lpnamer/problem_modifier/ProblemModifier.cpp
JasonMarechal25: Goes hand in hand with the next comment "You use a dedicated method candidateContributionIsNotNull above, why not here ?"
This function is the result of an _extract method_ and is wrongly named. It is missing the "direct" aspect of it.
The extract is more about naming the operation/test than code factorisation.
this function is now renamed

src/cpp/lpnamer/problem_modifier/ProblemModifier.cpp
JasonMarechal25: This part is now extracted in a new function _candidateContributionIndirectIsNotNull_

src/cpp/lpnamer/problem_modifier/StudyUpdateLinkCapacitiesStrategy.cpp
JasonMarechal25: Fixed in commit, thanks

src/cpp/lpnamer/problem_modifier/StudyUpdateLinkParameterStrategy.cpp
JasonMarechal25: Old habits...
Thanks

a-zakir (approved):

PR #454, opened 2022-05-10
Fix test not running properly when buiding outside repository

JasonMarechal25:
When building xpansion outside xpansion repository root some tests would fail trying to access test files.
The issue come from a wrong definition of working directory for these tests

### DIFF
--- a/tests/cpp/sensitivity/SensitivityInputReaderTest.cpp
+++ b/tests/cpp/sensitivity/SensitivityInputReaderTest.cpp
@@ -15,12 +15,7 @@ class SensitivityInputReaderTest : public ::testing::Test {
 
  protected:
   void SetUp() override {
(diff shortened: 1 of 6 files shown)


tests/cpp/sensitivity/SensitivityInputReaderTest.cpp
```diff
@@ -16,9 +16,9 @@ class SensitivityInputReaderTest : public ::testing::Test {
  protected:
   void SetUp() override {
 #if defined(WIN32) || defined(_WIN32)
```
tbittar: The `#if defined #else #endif` is not necessary anymore, same remark for all other files

tbittar (changes requested):
We may remove useless `#if defined ...`

tbittar (approved):

PR #461, opened 2022-06-09
Release/0.7.0

JasonMarechal25:


a-zakir (dismissed):

tbittar (dismissed):

tests/cpp/solvers_interface/test_basis.cpp
```diff
@@ -0,0 +1,56 @@
...
+void assert_basis_equality(SolverAbstract::Ptr expec_solver,
+                           SolverAbstract::Ptr current_solver) {
+  std::vector<int> expec_rstatus(expec_solver->get_nrows());
+  std::vector<int> expec_cstatus(expec_solver->get_ncols());
+  expec_solver->get_basis(expec_rstatus.data(), expec_cstatus.data());
+
+  std::vector<int> current_rstatus(current_solver->get_nrows());
+  std::vector<int> current_cstatus(current_solver->get_ncols());
+  current_solver->get_basis(current_rstatus.data(), current_cstatus.data());
+
```
flomnes, line 14: REQUIRE that `expec_solver->get_nrows()` be equal to `current_solver->get_nrows()` (resp `get_ncols`) ?

Otherwise the loops below would result in invalid reads.

tests/cpp/solvers_interface/test_basis.cpp
```diff
@@ -0,0 +1,56 @@
...
+  for (int i(0); i < expec_solver->get_nrows(); i++) {
+    REQUIRE(expec_rstatus[i] == current_rstatus[i]);
+  }
+  for (int i(0); i < expec_solver->get_ncols(); i++) {
+    REQUIRE(expec_cstatus[i] == current_cstatus[i]);
+  }
```
flomnes, line 20: Apply the remark above, or just use `operator==` for `std::vector<T>`. You may not get verbose output though.

tests/cpp/solvers_interface/test_basis.cpp
```diff
@@ -0,0 +1,56 @@
...
+
+        SolverAbstract::Ptr current_solver = factory.create_solver(solver_name);
+        current_solver->init();
+        current_solver->read_prob_mps(instance);
+        current_solver->read_basis(basis_file);
+
+        assert_basis_equality(expec_solver, current_solver);
+
+        expec_solver->free();
+        current_solver->free();
```
flomnes, line 52: Why not call `free` in method `SolverAbstract::~SolverAbstract` ? Since `SolverAbstract::Ptr` is a shared_ptr, method `free` for the derived class (XPRESS, CPLEX, CBC, etc.) would be called here.

You're at risk of memory leaks with the current implementation.

tests/cpp/solvers_interface/test_basis.cpp
tbittar: Yes this will be fixed in the develop branch, not directly on the release

tests/cpp/solvers_interface/test_basis.cpp
tbittar: Same as above

tests/cpp/solvers_interface/test_basis.cpp
tbittar: `free` is called manually at the end of `launch()` in `BendersSequential` and  `BendersMPI`. More investigation is required before applying your proposition.

tbittar (approved):

a-zakir (approved):

PR #467, opened 2022-06-17
fix: case null values in basis

a-zakir:


a-zakir:
issue is fixed in #469

PR #469, opened 2022-06-23
Fix/write coin basis

### LINKED ISSUE #465: Sensitivity analysis bug when tested with windows
Following error message, when testing the functionality on the studies of the "example" folder :

image

An idea would be to add systematically tests with these studies of the "example" folder on windows before each release.

tbittar:
Fix Cbc solver interface basis was not written/read correctly to enable warm start, closes #465


src/cpp/multisolver_interface/SolverXpress.cpp
```diff
@@ -119,7 +119,7 @@ void SolverXpress::write_prob_lp(const std::filesystem::path &filename) {
 }
 
 void SolverXpress::write_basis(const std::filesystem::path &filename) {
-  int status = XPRSwritebasis(_xprs, filename.string().c_str(), "");
+  int status = XPRSwritebasis(_xprs, filename.string().c_str(), "n");
```
JasonMarechal25: Can you extract the magic value "n" in a named constant please ?

src/cpp/sensitivity/AnalysisFunctions.cpp
```diff
@@ -31,21 +27,29 @@ SolverAbstract::Ptr get_sensitivity_problem(
...
+      sensitivity_problem = pb_modifier.changeProblem(nb_candidates);
+      break;
     }
 
     default:
       std::cerr << "Unrecognized Sensitivity Problem type" << std::endl;
       return nullptr;
   }
+
+  if (std::ifstream basis_file(input_data.basis_file_path); basis_file.good()) {
```
JasonMarechal25: If the file is not good nothing happens. Should we at least warn users or developers with log ?

JasonMarechal25 (changes requested):

src/cpp/sensitivity/AnalysisFunctions.cpp
tbittar: Yes, we can add a warning

src/cpp/multisolver_interface/SolverXpress.cpp
a-zakir: in doc it's a `const char* flags `

a-zakir (approved):

PR #481, opened 2022-07-13
manage case candidates with no capa files

### LINKED ISSUE #475: Existence of capacity directory should be optional
In studies without link profiles, there is no need to have a `capa` directory (meant to store the link profiles). However, even for such studies, Xpansion raises an error:
MicrosoftTeams-image

For such studies, no error should be raised.

Besides, the error message should print the path of the expected `capa` directory and not the candidate file.

a-zakir:


a-zakir:
Fixes #475

a-zakir:
fixes #475

JasonMarechal25 (approved):

PR #482, opened 2022-07-13
Fix/improve sensitivity log

### LINKED ISSUE #474: Recall optimal capacties when starting sensitivity analysis
When launching sensitivity analysis, it could be interesting to print the results of the Benders run (optimal investment capacities) so that the user can analyse at a glance the output of the sensitivity analysis with regard to the optimal capacities

tbittar:
Closes #474

### DIFF
--- a/src/cpp/sensitivity/SensitivityLogger.cpp
+++ b/src/cpp/sensitivity/SensitivityLogger.cpp
@@ -63,7 +63,8 @@ void SensitivityLogger::log_at_ending() {
   _stream << "--- Sensitivity study completed ---" << std::endl;
 }
 
-void SensitivityLogger::log_summary(const SensitivityInputData& input_data, const SensitivityOutputData& output_data) {
+void SensitivityLogger::log_summary(const SensitivityInputData& input_data,
+                                    const std::vector<SinglePbData>& pbs_data) {
   _stream << std::endl
           << "--- Sensitivity analysis summary "
           << "(epsilon = " << input_data.epsilon << EUROS << ") ---"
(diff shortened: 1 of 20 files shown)


a-zakir (approved):

src/cpp/sensitivity/SensitivityLogger.cpp
```diff
@@ -14,14 +14,47 @@ void SensitivityLogger::display_message(const std::string& msg) {
...
+  _stream << indent_1 << "Best capex = "
+          << xpansion::logger::commons::create_str_million_euros(benders_capex)
+          << MILLON_EUROS << std::endl
+          << std::endl;
+}
+
+void SensitivityLogger::log_benders_solution(
+    const std::map<std::string, double>& benders_solution) {
+  _stream << indent_1 << "Best investment solution = " << std::endl;
+  for (const auto& kvp : benders_solution) {
```
JasonMarechal25: You can use structured binding. You can even use better names than "key" and "value", like "candidate" and "investement" or what is proper in this case.
```suggestion
  for (const auto& [key, value] : benders_solution) {
      _stream << indent_1 << indent_1 << key << ": "
            << xpansion::logger::commons::create_str_mw(value) << MW
```

src/cpp/sensitivity/include/SensitivityILogger.h
```diff
@@ -3,17 +3,19 @@
 
 #include <string>
 
+#include "SensitivityInputReader.h"
```
JasonMarechal25, line 6: Why is this include necessary?

src/cpp/sensitivity/include/SensitivityWriter.h
```diff
@@ -4,6 +4,7 @@
 
 #include <filesystem>
 
+#include "SensitivityInputReader.h"
```
JasonMarechal25, line 7: Same as previously, why is the include necessary ?

src/cpp/sensitivity/include/SensitivityILogger.h
tbittar: Useful to know the SensitivityinputData structure, should we put it in a separate file ?

src/cpp/sensitivity/include/SensitivityWriter.h
tbittar: See answer above

src/cpp/sensitivity/SensitivityLogger.cpp
tbittar: Done

PR #483, opened 2022-07-19
Minor code corrections

JasonMarechal25:


src/cpp/benders/benders_core/include/common.h
```diff
@@ -136,15 +136,15 @@ struct BendersBaseOptions : public BaseOptions {
   explicit BendersBaseOptions(const BaseOptions &base_to_copy)
       : BaseOptions(base_to_copy) {}
 
-  int MAX_ITERATIONS;
+  unsigned int MAX_ITERATIONS = 1;
```
tbittar: I think MAX_ITERATIONS can take the value -1 for the case of an unlimited number of iterations, this should be checked

JasonMarechal25:
Deprecated

src/cpp/benders/benders_core/include/common.h
JasonMarechal25: I fixed the wrong way. There was an mismatch between signed and unsigned, I made this one unsigned, it should have been the other way.
I've set default value to -1 per your advice. In all case the reader would set it to -1 if the option is missing

tbittar (approved):

PR #486, opened 2022-07-25
ziplib adr

a-zakir:
This branch contains Adr to choose zip lib


conception/ziblib/choose-ziplib.md
```diff
@@ -0,0 +1,19 @@
+Choose Zip Library 
+===
+This Adr aim to select zip library that will be used to manage mps file in arhive.
```
flomnes:
```suggestion
This ADR aims to select the library that will be used to manage MPS files in ZIP archives.
```

conception/ziblib/choose-ziplib.md
```diff
@@ -0,0 +1,19 @@
...
+
+Status
+=== 
+libzip and it's wrapper libzippp : suspended
+
```
flomnes: Add minizip-ng

flomnes (commented):
`libzip` & `libzippp` were rejected because of the way they handle writes.

- Create a temporary copy of the archive `cp archive.zip archive.zip.<random suffix>`
- Write the new entry/entries to the temporary archive
- Replace the archive with the temporary archive `mv archive.zip.<random suffix> archive.zip`

This strategy ensures that a valid archive remains even in case the program is interrupted, `mv` being considered an atomic operation. However, it generates a lot of I/O, which is not suitable for us. This behavior cannot be disabled, meaning it is impossible to write entries straight to the archive.

`minizip-ng` on the other hand writes directly to the archive. Attention must be paid to call `mz_zip_writer_close` in case of user/system interruption. Signal handlers can be used, see e.g https://github.com/AntaresSimulatorTeam/Antares_Simulator/pull/827

flomnes (approved):

PR #487, opened 2022-07-25
[Archived] Feature/memory patch

JasonMarechal25:


docs/user-guide/get-started/launching-optimization.md
```diff
@@ -91,6 +91,13 @@ Show the Antares-Xpansion version.
 
 Show the Antares-Simulator version (used in the `antares` step).
 
+### `--construct_all_problems`
+
+Default value: `True`.
+
+
+`{True|False}`. Define is you want the benders step to construct all problems initially (True) or to rebuild them each iteration (False). Set to False to reduce maximum RAM usage at the cost of longer execution time (e.g. Running a large study on a personal computer.)
```
tbittar: is you want -> if you want
rebuild them at each iteration

src/cpp/benders/benders_core/BendersBase.cpp
```diff
@@ -744,10 +815,14 @@ WorkerMasterPtr BendersBase::get_master() const { return _master; }
...
-      SubproblemWeight(_data.nsubproblem, kvp.first), _options.SOLVER_NAME,
-      _options.LOG_LEVEL, log_name());
+  subproblem_map[kvp.first] = makeSubproblemWorker(kvp);
+}
+std::shared_ptr<SubproblemWorker> BendersBase::makeSubproblemWorker(
+    const std::pair<std::string, VariableMap> &kvp) const {
+  return std::make_shared<SubproblemWorker>(
+        kvp.second, GetSubproblemPath(kvp.first),
+      SubproblemWeight(_data.nsubproblem, kvp.first), Options().SOLVER_NAME,
+      Options().LOG_LEVEL, log_name());
```
tbittar, line 825: Why do we use `Options()` instaed of `options_` ?

src/cpp/lpnamer/model/Problem.h
```diff
@@ -103,6 +103,7 @@ class Problem: public SolverAbstract
   void set_simplex_iter(int iter) override { solver_abstract_->set_simplex_iter(iter);}
   void write_basis(const std::filesystem::path &filename) override { solver_abstract_->write_basis(filename); }
   void read_basis(const std::filesystem::path &filename) override { solver_abstract_->read_basis(filename); }
+  void SetBasis(std::vector<int> rstatus, std::vector<int> cstatus) override { solver_abstract_->SetBasis(rstatus, cstatus); }
```
tbittar, line 106: Why do you chose a different formatting in the function name compared to the other functions of the class ? (Because the input data type is different ?)

tbittar (commented):
Should we add tests for BendersMPI similarly as for BendersSequential ?

src/cpp/benders/benders_core/BendersBase.cpp
JasonMarechal25: It makes no difference, Options() is just a getter on options_

src/cpp/lpnamer/model/Problem.h
JasonMarechal25: That's a difficult point.
On one hand we should keep formatting homogeneous in files, on another hand is we don't take the habits to use our chosen format we will never do.

PR #494, opened 2022-08-03
Fix/488 fix algorithm not stopping at max iteration

### LINKED ISSUE #488: The algorithm stops at `max_iteration` + 1 iterations
When the algorithm stops with the `max_iteration` criterion, it stops after `max_iteration` + 1 iterations. We expect the algorithm to stop at the end of iteration `max_iteration`.

JasonMarechal25:


src/CMakeLists.txt
```diff
@@ -36,6 +36,7 @@ find_program(PYINSTALLER "pyinstaller")
 
 #if pyinstaller available add compiled version of python scripts
 if (PYINSTALLER)
+    install(FILES ${CMAKE_CURRENT_SOURCE_DIR}/python/config.yaml DESTINATION . )
```
a-zakir: why copying this file? in `if else` condition the file is integrated in `antares-xpansion-launcher` exe

JasonMarechal25: I had an issue where I could not start antares-launcher in my install directory. It was looking for this file.

tbittar (approved):

src/CMakeLists.txt
JasonMarechal25: J'ai retiré, j'ai probablement une config local qui rendait ça nécessaire (pour moi).

PR #511, opened 2022-08-31
Force storenewset to true, fix chronicles_checker

tbittar:
Two fixes :
- Fix missing `.shape` in chronicle checker
- Force `storenewset` to `true` in order to always have the output of Antares scenario builder and be able to choose the right chronicle for the right MC year
- Fix some typos and improve a warning message

### DIFF
--- a/src/cpp/lpnamer/model/ChronicleMapProvider.cpp
+++ b/src/cpp/lpnamer/model/ChronicleMapProvider.cpp
@@ -2,24 +2,31 @@
 // Created by marechaljas on 29/04/2022.
 //
 
-#include "ChronicleMapReader.h"
-#include <utility>
-#include <fstream>
-#include <filesystem>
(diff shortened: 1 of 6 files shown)


src/cpp/lpnamer/model/ChronicleMapProvider.cpp
```diff
@@ -2,24 +2,31 @@
...
 //
 
-#include "ChronicleMapReader.h"
-#include <utility>
-#include <fstream>
-#include <filesystem>
-#include <gtest/gtest.h>
 #include "ChronicleMapProvider.h"
+
+#include <gtest/gtest.h>
```
JasonMarechal25: Please remove gtest include (yes it was already there but shouldn't)

a-zakir (approved):

PR #515, opened 2022-09-01
Logger for Problem_Generation 

### LINKED ISSUE #509: Logger for problem_generation
For now log messages from problem_generation step are  catched only by the python orchestrator and redirected to a file.
I think it would be interesting to create a logger in problem_generation's c++ lib.
or maybe a unique logger for all xpansion.

a-zakir:
closes #509


JasonMarechal25:
We already have a dependency to glog. Would it be a good solution to generalize usage of glog library and use it for your logger ?

JasonMarechal25 (approved):

PR #520, opened 2022-09-12
Improve NTC error msg and link profile parsing

### LINKED ISSUE #519: Improve error messages when NTC and candidate column number mismatch
Currently when there is a mismatch between a candidate link-profile and the corresponding NTC profile in the Antares study, the error message is `antares_xpansion.chronicles_checker.NTC_And_Candidate_Mismatch`

There is no information on the candidate profile file that is involved nor the shape of the two files that must match, this information should be included in the error message to ease debugging.

tbittar:
Closes #519 and improve parsing of link profile file (now any number of spaces or tabs is allowed between columns)

### DIFF
--- a/src/python/antares_xpansion/candidates_reader.py
+++ b/src/python/antares_xpansion/candidates_reader.py
@@ -4,7 +4,7 @@ from typing import List
 
 import numpy as np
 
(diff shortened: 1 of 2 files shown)


tbittar:
Closing PR as changing base branch to current release

PR #527, opened 2022-09-14
Feature/download deps assets

JasonMarechal25:
The goal of this PR is to enable downloading dependencies assets to prevent having to build them and save time

- [x] Download ubuntu Antares simulator asset
- [ ] Download windows Antares simulator asset
- [x] Download centos Antares simulator asset
- [ ] Distinguish between xpress and no xpress


docs/developer-guide/install_from_sources/3-Build.md
```diff
@@ -30,18 +30,19 @@ git submodule update --init antares-deps
...
-|`BUILD_UI`|`OFF`|Enable UI build.|
-|`ALLOW_RUN_AS_ROOT`|`OFF`|allow mpi to run as root for centOs docker.|
+| Option                     | Default                                  | Description                                                                  |
+|:---------------------------|------------------------------------------|------------------------------------------------------------------------------|
+| `CMAKE_BUILD_TYPE`         | `Release`                                | Define build type. Available values are `Release` and `Debug`.               |
+| `USE_SEQUENTIAL`           | `ON`                                     | Enable build of sequential executable.                                       |
+| `USE_MPI`                  | `OFF`                                    | Enable build of bendersmpi executable.                                       |
+| `DBUILD_antares_solver`    | `OFF`                                    | Enable build of antares-solver.                                              |
+| `DDOWNLOAD_antares_solver` | `ON`                                     | Enable download of antares-solver.                                           |
```
flomnes, line 39:
```suggestion
| `BUILD_antares_solver`    | `OFF`                                    | Enable build of antares-solver.                                              |
| `DOWNLOAD_antares_solver` | `ON`                                     | Enable download of antares-solver.                                           |
```

flomnes: Are those two options compatible ? What happens if they're both set to ON ?

docs/developer-guide/install_from_sources/3-Build.md
```diff
@@ -57,6 +58,18 @@ Additionnal options for Xpress use:
...
 
+### Antares solver dependency
+You may have noticed two options related to antares-solver: _BUILD_antares_solver_ and _DOWNLOAD_antares_solver_
+
+In case cmake can't find _antares-solver_ package it will check if _DOWNLOAD_antares_solver_ is ON. In this case it will try 
+to download the asset from [Antares Simulator repository](https://github.com/AntaresSimulatorTeam/Antares_Simulator/releases).
+The asset downloaded is always a release version. The version number is the one set in antares-version.json at the root of XPansion repository
+
+In case the package is still not found (download failed, or you didn't enable it) and if _BUILD_antares_solver_ is ON cmake will 
+fetch [Antares Simulator repository](https://github.com/AntaresSimulatorTeam/Antares_Simulator/) and build it. The build type (debug|release)
```
flomnes, line 69:
```suggestion
In case the package is still not found (download failed, or you didn't enable it) and if _BUILD_antares_solver_ is ON cmake will 
fetch [Antares Simulator source files](https://github.com/AntaresSimulatorTeam/Antares_Simulator/) and build the project. The build type (debug|release)
```

docs/developer-guide/install_from_sources/3-Build.md
```diff
@@ -57,6 +58,18 @@ Additionnal options for Xpress use:
...
+depends on the build chosen for Xpansion. To build Antares Simulator you will need to have its dependency ortool
+available to cmake find_package()
```
flomnes, line 71:
```suggestion
depends on the build chosen for Xpansion. To build Antares Simulator you will need to have its dependency OR-Tools, which is available using CMake's find_package(). You may also provide an install directory in the `CMAKE_PREFIX_PATH`.
```

cmake/dependencies/antares-deps.cmake
```diff
@@ -0,0 +1,48 @@
...
+    if (DOWNLOAD_DEPS)
+        message(STATUS "Downloading antares deps asset")
+        if (MSVC)
+            set(PLATFORM_TYPE windows-latest-Release.zip)
+        else()
+            execute_process(COMMAND bash "-c" "lsb_release -d | awk -F\"\t\" '{print $2}'" OUTPUT_VARIABLE SYSTEM_FLAVOUR)
+            string(FIND ${SYSTEM_FLAVOUR} "Ubuntu" UBUNTU_FOUND)
+            if (UBUNTU_FOUND EQUAL -1)
+                message(STATUS "Will download Antares solver CentOS assets")
+                set(PLATFORM_TYPE centos7-Release.tar.gz)
```
flomnes, line 29: What happens in CentOS 8 ? Or Ubuntu 22.04 ?

cmake/dependencies/antares-deps.cmake
```diff
@@ -0,0 +1,48 @@
...
+        )
+        execute_process(COMMAND mkdir -p ${DEPS_INSTALL_DIR})
+        execute_process(
+                COMMAND tar -xvf rte-antares-deps-${PLATFORM_TYPE} -C ${DEPS_INSTALL_DIR} --strip-components=1
+        )
+        execute_process(
+                COMMAND rm -rf rte-antares-deps-${PLATFORM_TYPE}
+        )
+    else()
+        message(WARNING "One or more dependency is missing but DOWNLOAD_DEPS is OFF")
```
flomnes, line 46: Is everything available on Windows ? wget/tar especially

Consider using [CMake's file download](https://cmake.org/cmake/help/latest/command/file.html#transfer) for better compatibility. See also [archiving](https://cmake.org/cmake/help/latest/command/file.html#archiving) about decompression (requires CMake 3.18+ though)

cmake/dependencies/antares-solver.cmake
```diff
@@ -0,0 +1,64 @@
+#####################
+##  Antares solver  ##
+#####################
+message("DEPENDENCY ANTARES " ${BUILD_antares_solver})
+message("DEPENDENCY ANTARES " ${DOWNLOAD_antares_solver})
```
flomnes, line 5:
```suggestion
message("ANTARES SIMULATOR BUILD " ${BUILD_antares_solver})
message("ANTARES SIMULATOR DOWNLOAD " ${DOWNLOAD_antares_solver})
```

CMakeLists.txt
```diff
@@ -167,26 +167,31 @@ sbeParseJson(antares_version jsonContent)
...
 SET(ANTARES_VERSION ${antares_version.antares_version_executable})
 SET(ANTARES_VERSION_TAG ${antares_version.antares_version})
+SET(ANTARES_DEPS_VERSION ${antares_version.antares_deps_version})
 
 sbeClearJson(antares_version)
 
 # Build dependencies not handled by antares-deps repository
-option(BUILD_antares_solver "Build antares solver" ON)
+option(BUILD_antares_solver "Build antares solver" OFF)
+option(DOWNLOAD_antares_solver "Download antares solver" ON)
```
flomnes, line 176: I suggest using CMAKE_DEPENDENT_OPTION, since these options are exclusive (it's silly to build AND download antares-solver)

cmake/dependencies/antares-solver.cmake
```diff
@@ -0,0 +1,64 @@
...
+            message(${DEPS_INSTALL_DIR}/..)
+            if (MSVC)
+            message(STATUS "Will download Antares solver windows assets")
+                set(PLATFORM_TYPE installer-64bits.zip)
+            else()
+                execute_process(COMMAND bash "-c" "lsb_release -d | awk -F\"\t\" '{print $2}'" OUTPUT_VARIABLE SYSTEM_FLAVOUR)
+                string(FIND ${SYSTEM_FLAVOUR} "Ubuntu" UBUNTU_FOUND)
+                if (UBUNTU_FOUND EQUAL -1)
+                    message(STATUS "Will download Antares solver CentOS assets")
+                    set(PLATFORM_TYPE CentOS-7.9.2009-ortools-xpress.tar.gz)
```
flomnes, line 19: Careful, this asset requires XPRESS. By using it, your Xpansion build will also depend on it.

cmake/dependencies/antares-solver.cmake
```diff
@@ -0,0 +1,64 @@
...
+                else()
+                    message(STATUS "Will download Antares solver Ubuntu assets")
+                    set(PLATFORM_TYPE ubuntu-20.04.tar.gz)
+                endif()
+            endif()
+            execute_process(COMMAND rm -rf antares-${ANTARES_VERSION_TAG}-${PLATFORM_TYPE}) #Remove old version, just in case
+            execute_process(COMMAND
+                    wget https://github.com/AntaresSimulatorTeam/Antares_Simulator/releases/download/v${ANTARES_VERSION_TAG}/antares-${ANTARES_VERSION_TAG}-${PLATFORM_TYPE}
+                    )
```
flomnes, line 28: - `wget -O output` overwrites `output`
- Naming is not uniform rte-antares (Windows), antares-*-Ubuntu, antares-*-CentOS, etc. It seems that l. 25 only works in Ubuntu & CentOS.

flomnes (changes requested):
Removing antares-deps as a submodule is a great simplification. It should be done in Simulator as well since it's confusing.

Building antares-solver without the GUI isn't very long. It's the build of OR-Tools and its deps (zlib, protobuf, absl, etc.) that takes time. Make sure to mention it in the docs :
- Download antares-solver **and its dependencies**
- Build antares-solver **and its dependencies**

Make sure to test in Windows, since I have doubts about compatibility with Windows.

Include this new option in the CI ?

cmake/dependencies/antares-deps.cmake
JasonMarechal25: Thank you, will look into this

cmake/dependencies/antares-deps.cmake
JasonMarechal25: As of now those platforms are unsupported by both simulator and xpansion it's not issue for the moment. However if we come to support different variant of ubuntu or centOS (well more plausible ubuntu than centOS) then we will need to refine the regexp

cmake/dependencies/antares-deps.cmake
JasonMarechal25: Using cmake file API could be a nice way to ensure compatibility between the systems. Unfortunately for ARCHIVE_EXTRACT cmake 3.18 is required. Ubuntu 20.04 comes with 3.16 and centOS 7 with 3.17. We would need to manually install an up to date version. This begets a new discussion

cmake/dependencies/antares-solver.cmake
JasonMarechal25: Indeed didn't see it

cmake/dependencies/antares-deps.cmake
flomnes: We have the same problem with  newer versions of OR-Tools, it requires CMake 3.18+, which isn't available for these platforms.

JasonMarechal25:
With the current direction of the projects this PR is not needed anymore.
We will probably go for a mono repo and maybe fuse simulator and xpansion.
We may also kill antares deps and integrate the meaningful content directly in projects

PR #540, opened 2022-09-23
Properly set missing general data value to expected xpansion values

### LINKED ISSUE #537: Modify `generaldata.ini` correctly even if the field is initially missing
In `general_data_processor.py`, the `generaldata.ini` is modified to fit Xpansion needs. However if some fields are missing (and considered set to their default value by Simulator) they are not modified as expected in xpansion, leading to an error or an unexpected behavior of the simulation.

If the field is missing, we expect that xpansion creates it with the value it expects.

JasonMarechal25:


a-zakir:
**test_no_changes_in_empty_file** is deprecated since generadata.ini must contains minimal sections and attribute.

tbittar (approved):

PR #545, opened 2022-10-06
change solver basis format to normal

a-zakir:
avoid seg fault due to undefined behavior in windows (writeBasis function of ClpSimplex)


JasonMarechal25 (approved):

JasonMarechal25:
@a-zakir Can you write an ADR explaining the issue and put a comment refferencing the adr next to writeBasis usage or declaration please.
This way we won't lose the information on why 0 is used

PR #561, opened 2022-10-26
Update default values and fix docs

### LINKED ISSUE #560: Choose good default values for in-out stabilization
Default values of the parameters `separation_parameter`and  `relaxed_optimality_gap` should be tuned so that the in-out stabilization is efficient. The docs should be updated as well.

tbittar:
Closes #560


src/cpp/benders/benders_core/BendersBase.cpp
```diff
@@ -525,16 +526,10 @@ LogData BendersBase::build_log_data_from_data() const {
 LogData BendersBase::FinalLogData() const {
   LogData result;
 
-  result.it = _data.it + iterations_before_resume;
   result.best_it = _data.best_it + iterations_before_resume;
-  result.lb = _data.lb;
-  result.best_ub = _data.best_ub;
-  result.x_out = _data.x_in;
```
a-zakir, line 532: why ?

a-zakir (approved):

src/cpp/benders/benders_core/BendersBase.cpp
tbittar: Data is unused

PR #567, opened 2022-11-14
benders by batch

a-zakir:
Implementing benders by batch.
For now the algo works under condition :
- master = relaxed
- without stabilization
- absolute_gap must be used as stop criterion.

About batch : 
-  batches are constructed with a dispatch strategy using the id of subproblems (e.g. given sub problems (p_id):   p_1, p_2, p_3, p_4, p_5 and batch_size=2 then batch1 ={p_1, p_2}, batch2 = {p_3,p_4} and batch3 = {p_5})
- a cyclic strategy is used to get a permutation of batches at each iteration (if batch1 and batch2 are solved in iteration i  then in i+1 iteration batch3 will be the first solved then batch1 etc.. ). Random strategy is also implemented but not tested (unavailable for user).


src/cpp/benders/benders_by_batch/BatchCollection.cpp
```diff
@@ -0,0 +1,34 @@
...
+#include <cmath>
+BatchCollection::BatchCollection(
+    const std::vector<std::string>& sub_problem_names, size_t batch_size) {
+  const auto sub_problems_number = sub_problem_names.size();
+
+  if (batch_size > sub_problems_number) {
+    throw BatchCollection::InvalidSizeOfSubPrblemCollection(
+        batch_size, sub_problems_number);
+  }
+  if (batch_size == sub_problems_number) {
```
tbittar: I think there is no need to distinguish the case with batch_size = subproblems_number, lines 12 to 14 can be removed

src/cpp/benders/benders_by_batch/BendersByBatch.cpp
```diff
@@ -0,0 +1,204 @@
...
+  if (is_initial_relaxation_requested()) {
+    _logger->LogAtInitialRelaxation();
+    DeactivateIntegrityConstraints();
+    SetDataPreRelaxation();
+  }
+
+  auto batch_size = Options().BATCH_SIZE == 0 ? GetSubProblemNames().size()
+                                              : Options().BATCH_SIZE;
+
+  double remaining_epsilon = AbsoluteGap();
```
tbittar: Useless as this is initialized line 86

src/cpp/benders/benders_by_batch/include/BatchCollection.h
```diff
@@ -0,0 +1,35 @@
...
+};
+class BatchCollection {
+ private:
+  std::vector<Batch> batch_collections_;
+  unsigned number_of_batch_;
+
+ public:
+  BatchCollection(const std::vector<std::string>& sub_problem_names,
+                  size_t size_of_sub_problems_collection);
+  class InvalidSizeOfSubPrblemCollection : public std::runtime_error {
```
tbittar: typo SubproblemCollection

src/cpp/benders/benders_by_batch/include/BatchCollection.h
```diff
@@ -0,0 +1,35 @@
...
+   public:
+    explicit InvalidSizeOfSubPrblemCollection(size_t size,
```
tbittar: typo SubproblemCollection

src/cpp/benders/benders_core/BendersBase.cpp
```diff
@@ -430,7 +416,7 @@ void BendersBase::getSubproblemCut(
  *
  */
 void BendersBase::compute_cut(AllCutPackage const &all_package) {
-  _data.ub = 0;
+  // _data.ub = 0;
```
tbittar: Remove the commented code and check if this does not break Benders sequential / MPI (as _data.ub = 0 is in build_cut)

src/cpp/benders/logger/User.cpp
```diff
@@ -85,14 +85,17 @@ void User::restart_best_iterations_infos(const LogData &best_iteration_data) {
...
 }
 
 void User::LogAtSwitchToInteger() {
   _stream << "--- Relaxed gap reached, switch master formulation to integer"
           << std::endl;
 }
-
+void User::number_of_sub_problem_resolved(int number) {
+  _stream << indent_1
+          << "cumulative number of call to solver (only for subproblems): "
```
tbittar: "cumulative number of subproblem resolutions" to be more concise ?

src/python/antares_xpansion/config_loader.py
```diff
@@ -313,6 +316,17 @@ def get_separation(self):
 
         return float(separation_parameter_str)
 
+    def get_batch_size(self):
+        """
+        return the separation parameter read from the settings file
```
tbittar: Update or remove comment

src/cpp/benders/benders_by_batch/BatchCollection.cpp
```diff
@@ -0,0 +1,34 @@
+#include "BatchCollection.h"
+
+#include <cmath>
+BatchCollection::BatchCollection(
+    const std::vector<std::string>& sub_problem_names, size_t batch_size) {
+  const auto sub_problems_number = sub_problem_names.size();
+
+  if (batch_size > sub_problems_number) {
```
tbittar, line 8: As the Python part can only check if batch_size is integer >= 0, I think it is natural to set batch_size = sub_problems_number here so that we avoid raising an error after the user has run antares and problem_generation and spend already some time in the run of the study.

src/python/antares_xpansion/xpansionConfig.py
```diff
@@ -173,6 +177,7 @@ def _set_default_options(self):
             OptimisationKeys.csv_name_key(): self.csv_name_default_value(),
             OptimisationKeys.bound_alpha_key(): self.bound_alpha_default_value(),
             OptimisationKeys.separation_key(): self.separation_default_value(),
+            OptimisationKeys.batch_size_key(): self.batch_size_value(),
```
tbittar: rename to batch_size_default_value for coherence ?

tests/cpp/benders/BendersByBatchTest.cpp
```diff
@@ -0,0 +1,87 @@
...
+  const std::vector<std::string> sub_problems_name_list = {"P1", "P2", "P3"};
+  const std::vector<std::string> sub_problems_name_list_4 = {"P1", "P2", "P3",
+                                                             "P4"};
+  const std::vector<std::string> sub_problems_name_list_5 = {"P1", "P2", "P3",
+                                                             "P4", "P5"};
+
+ protected:
+};
+
+TEST_F(BatchCollectionTest,
```
tbittar, line 19: This test may be modified if we decide not to raise an exception in this case

tests/cpp/benders/BendersByBatchTest.cpp
```diff
@@ -0,0 +1,87 @@
...
+  }
+  ASSERT_EQ(sum_batch_size, sub_problems_name_list.size());
+}
+TEST_F(BatchCollectionTest,
+       IfNumberOFSubProblemIs3AndBatchSize2ThenNumberOfBatchIs2) {
+  auto batch_collection = BatchCollection(sub_problems_name_list, 2);
+
+  ASSERT_EQ(batch_collection.NumberOfBatch(), 2);
+}
+TEST_F(BatchCollectionTest,
```
tbittar, line 55: I think the tests of line 61 makes all the necessary checks and that tests from lines 32 to 55 are not useful anymore. Indeed if we check that the batch are constituted as expected then automatically the number of batch, the batch sizes and so on is correct. Maybe it is worth to launch the test of line 61 with different instances of sub_problems_name_list to cover all cases

tests/cpp/benders/BendersByBatchTest.cpp
```diff
@@ -0,0 +1,87 @@
...
+  const std::vector<std::string> expected_vec_batch0_names = {"P1", "P2"};
+  const std::vector<std::string> expected_vec_batch1_names = {"P3", "P4"};
+  const std::vector<std::string> expected_vec_batch2_names = {"P5"};
+  ASSERT_EQ(expected_vec_batch0_names, vec_batch0);
+  ASSERT_EQ(expected_vec_batch1_names, vec_batch1);
+  ASSERT_EQ(expected_vec_batch2_names, vec_batch2);
+}
+
+class RandomBatchShufflerTest : public ::testing::Test {};
+TEST_F(RandomBatchShufflerTest, GetRandomBatchPermutation) {
```
tbittar: This check is weird, I would remove it, and add a check for the cyclic permutation instead

tests/cpp/logger/logger_test.cpp
```diff
@@ -593,6 +593,8 @@ class SimpleLoggerMock : public ILogger {
   void LogAtInitialRelaxation() { _initialRelaxationCall = true; }
 
   void LogAtSwitchToInteger() { _switchToIntegerCall = true; }
+  void number_of_sub_problem_resolved(int number) {  //
```
tbittar, line 596: Is the // useful ?

tbittar (dismissed):

src/cpp/benders/benders_by_batch/BatchCollection.cpp
a-zakir: even if Benders is used in the antares-pb generation chain, it remains a separate executable. There is no guarantee that between Pb-generation and benders the options file is not corrupted (e.g. batch_size is really <= sub_problems_number)

tests/cpp/benders/BendersByBatchTest.cpp
a-zakir: there is a very low probability that this test will fail, it tests that GetRandomBatchOrder will shuffle an ascending ordered vector and that the result must be different from the input vector.

tests/cpp/logger/logger_test.cpp
a-zakir: yes to avoid smell code (empty function...)

tests/cpp/benders/BendersByBatchTest.cpp
a-zakir: theses tests were implemented in development process, they are just special cases (batch_size = 1, bactch_size and sub_num = even num, etc...). I think we can keep them.

JasonMarechal25 (approved):

PR #582, opened 2022-12-07
delete print

a-zakir:


tbittar:
I think that i have already removed this print in the release branch, so I do not know if this PR is required

a-zakir:
> I think that i have already removed this print in the release branch, so I do not know if this PR is required

Indeed, the patch was merged but it was lost, probably due to a later bad integration.

tbittar (approved):

PR #585, opened 2022-12-07
allow to work with  unzipped mps after Pb Gen

a-zakir:
This Pr will give the choice to zip the mps or to work with uncompressed files after PbGen


.github/workflows/build_centos7.yml
```diff
@@ -144,4 +144,4 @@ jobs:
           export LD_LIBRARY_PATH=LD_LIBRARY_PATH:${{ env.XPRS_LIB_Path_CONTAINER }}
           export XPRESS=${{ env.XPRESS_CONTAINER }}
           cd _build
-          ctest3 -C Release --output-on-failure -L "unit|benders|lpnamer|medium"
```
JasonMarechal25: Why are "medium" test removed ?

a-zakir: temporarily; to get the centos asset because otherwise the ci fails

src/cpp/helpers/ArchiveReader.h
```diff
@@ -1,5 +1,6 @@
 #ifndef _ARCHIVEREADER_H
 #define _ARCHIVEREADER_H
+#include <functional>
```
tbittar: What's the use of this new include ?

tbittar (approved):

PR #621, opened 2023-02-16
Feature/benders-by-batch-with-mpi

a-zakir:
enhance benders-by-batch with mpi parallelization


src/cpp/benders/benders_by_batch/BendersByBatch.cpp
```diff
@@ -117,42 +167,34 @@ void BendersByBatch::build_cut(
...
   *sum = 0;
-  // With gcc9 there was no parallelisation when iterating on the map directly
-  // so with project it in a vector
   std::vector<std::pair<std::string, SubproblemWorkerPtr>> nameAndWorkers;
-  nameAndWorkers.reserve(batch_sub_problems.size());
-  auto sub_pblm_map = GetSubProblemMap();
-  for (const auto &name : batch_sub_problems) {
-    nameAndWorkers.emplace_back(name, sub_pblm_map[name]);
+  const auto &sub_pblm_map = GetSubProblemMap();
+  std::copy_if(
```
tbittar: Can we get rid of the `copy_if` block as it was only useful to be able to use `for_each` in `BendersSequential`. I think the following `for` loop can directly iterate over the `sub_pblm_map`, with the appropriate `if` condition somewhere

src/cpp/benders/benders_by_batch/BendersByBatch.cpp
```diff
@@ -38,75 +70,93 @@ void BendersByBatch::run() {
...
+    BroadCast(random_batch_permutation_.data(),
+              random_batch_permutation_.size(), rank_0);
     batch_counter = 0;
 
     while (batch_counter < number_of_batch) {
       current_batch_id = random_batch_permutation_[batch_counter];
-      const auto &batch = batch_collection.GetBatchFromId(current_batch_id);
+      const auto &batch = batch_collection_.GetBatchFromId(current_batch_id);
       const auto &batch_sub_problems = batch.sub_problem_names;
       double sum = 0;
```
tbittar: Rename all occurrences by `single_proc_subpb_costs_gap` (or any better meaningful ideas ?)

src/cpp/benders/benders_by_batch/BendersByBatch.cpp
```diff
@@ -38,75 +70,93 @@ void BendersByBatch::run() {
...
-      build_cut(batch_sub_problems, &sum);
-      number_of_sub_problem_resolved += batch_sub_problems.size();
-      remaining_epsilon -= sum;
-      UpdateTrace();
-
-      if (remaining_epsilon > 0)
+      double result = 0;
```
tbittar: Rename all occurences by `subpb_costs_gap`

src/cpp/benders/benders_by_batch/BendersByBatch.cpp
```diff
@@ -38,75 +70,93 @@ void BendersByBatch::run() {
...
-  for (const auto &sub_problem_name : batch_sub_problems) {
-    auto sub_problem_data = subproblem_data_map[sub_problem_name];
-    SetSubproblemCost(GetSubproblemCost() + sub_problem_data.subproblem_cost);
+  for (const auto &subproblem_map : gathered_subproblem_map) {
+    for (auto &&[_, subproblem_data] : subproblem_map) {
+      SetSubproblemCost(GetSubproblemCost() + subproblem_data.subproblem_cost);
+    }
   }
 
-  SetSubproblemTimers(timer.elapsed());
```
tbittar, line 107: Where is the timer now ?

src/cpp/benders/benders_by_batch/include/BatchCollection.h
```diff
@@ -8,22 +9,48 @@
...
+  friend class boost::serialization::access;
+  template <class Archive>
+  void serialize(Archive &ar, const unsigned int version) {
+    ar &sub_problem_names;
+    ar &id;
+  }
 };
 class BatchCollection {
  private:
+  std::vector<std::string> sub_problem_names_;
```
tbittar, line 21: Do we really need to keep the `sub_problem_names_` as attributes here, as they are stored again in the vector of `Batch` ?

src/cpp/benders/benders_by_batch/include/BatchCollection.h
```diff
@@ -8,22 +9,48 @@
...
+  size_t batch_size_;
   std::vector<Batch> batch_collections_;
   unsigned number_of_batch_;
+  Logger logger_;
 
  public:
-  BatchCollection(const std::vector<std::string>& sub_problem_names,
+  BatchCollection() = default;
+  BatchCollection(const std::vector<std::string> &sub_problem_names,
                   size_t size_of_sub_problems_collection, Logger logger);
```
tbittar: `size_of_sub_problems_collection` should be renamed `batch_size` here

tbittar (changes requested):
We should update the docs with the name of the new tests

src/cpp/benders/benders_by_batch/include/BatchCollection.h
a-zakir: this attribute is used to build batches see BuildBatches method, otherwise batches must be built inside the ctor...

src/cpp/benders/benders_by_batch/BendersByBatch.cpp
a-zakir: updated

tbittar (approved):

PR #626, opened 2023-02-24
Sensitivity works with archive files

### LINKED ISSUE #624: Sensitivity analysis is broken
The outputs of Xpansion 1.0 are stored in a zip archive. However, the sensitivity analysis expects a folder structure, which causes the feature to fail.

JasonMarechal25:


src/python/antares_xpansion/sensitivity_driver.py
```diff
@@ -24,13 +25,16 @@ def launch(
...
+            xpansion_simulation_output
     ):
         """
         launch sensitivity analysis
         """
         self.simulation_output_path = self._get_simulation_output_path(
             simulation_output_path
         )
+        with zipfile.ZipFile(self.simulation_output_path, 'r') as output_zip:
+            output_zip.extractall(xpansion_simulation_output)
```
a-zakir, line 39: must extract all ?
what about the final sensitivity outputs, will they be compressed with other xpansion outputs

tbittar:
Closes #624

PR #628, opened 2023-03-01
remove deprecated code

a-zakir:
remove deprecated code that zip mps after Pb gen


src/cpp/lpnamer/problem_modifier/LinkProblemsGenerator.cpp
```diff
@@ -29,26 +29,20 @@
 
 void LinkProblemsGenerator::treat(
     const std::string &problem_name, Couplings &couplings,
-    std::shared_ptr<IProblemWriter> writer,
```
JasonMarechal25, line 32: Please keep the usage of a dedicated interface.
You can implement a "MPSFileWriter" to use instead of ArchiveProblemWriter and just call "in_prblm->write_prob_mps(lp_mps_name);" in "Write_problem(...)"

It may seems overkill but we just had multiple case of alternating between Zip or mps and may pass by memory in the future.

src/cpp/lpnamer/problem_modifier/LinkProblemsGenerator.cpp
```diff
@@ -67,47 +61,22 @@ void LinkProblemsGenerator::treat(
...
-                  auto problem_variables_from_zip_adapter =
-                      std::make_shared<ProblemVariablesZipAdapter>(
-                          reader, mps, _links, logger_);
-                  treat(mps._problem_mps, couplings, writer, adapter,
-                        problem_variables_from_zip_adapter);
-                });
+  in_prblm->write_prob_mps(lp_mps_name);
 }
 
 void LinkProblemsGenerator::treatloop_files(
```
JasonMarechal25: Since "treatloop" has been removed you can rename "treatloop_files" to "treatloop"

JasonMarechal25 (changes requested):

JasonMarechal25 (approved):

PR #635, opened 2023-03-15
relative gap as stopping criterion in benders by batch 

a-zakir:
implementation of relative gap criterion in benders by batch


tbittar:
This PR also handles the case `master = integer` in the Benders by batch with in-out stabilization.

tbittar (approved):

PR #640, opened 2023-03-21
log visiblity

a-zakir:
print log location: path, file, line and function


src/cpp/benders/benders_core/BendersBase.cpp
```diff
@@ -249,18 +250,23 @@ void BendersBase::ResetDataPostRelaxation() {
...
-    LOG(INFO) << "Master status is " << _data.master_status << std::endl;
-    throw InvalidSolverStatusException("Master status is " +
-                                       std::to_string(_data.master_status));
+    std::ostringstream msg;
+    msg << "Master status is " << _data.master_status << std::endl;
+    _logger->display_message(msg.str());
+    msg.str("");
+    auto log_location = LOGLOCATION;
+    msg << "Master status is " + std::to_string(_data.master_status);
+    _logger->display_message(log_location + msg.str());
```
tbittar, line 256: Master status seems to be printed twice

src/cpp/xpansion_interfaces/LogUtils.h
```diff
@@ -0,0 +1,24 @@
+#pragma once
+#include <stdexcept>
+#include <string>
+
+inline std::string LogLocationToStr(int line, const char* file,
+                                    const char* func) {
+  return std::string("This is line ") + std::to_string(line) + " of file " +
+         file + " (function " + func + ")\n";
```
flomnes, line 8: `"Function " + func + "(" + file + ":" + std::to_string(line) + ")"` ?

src/cpp/benders/benders_core/BendersBase.cpp
```diff
@@ -249,18 +250,20 @@ void BendersBase::ResetDataPostRelaxation() {
...
-      std::stringstream stream;
+      std::ostringstream stream;
+      auto log_location = LOGLOCATION;
       stream << "Subproblem " << subproblem_name << " status is "
-             << subproblemData.lpstatus;
-      LOG(INFO) << stream.str() << std::endl;
-
-      throw InvalidSolverStatusException(stream.str());
+             << subproblemData.lpstatus << std::endl;
+      _logger->display_message(log_location + stream.str());
```
flomnes, line 265: It would be nice if `_logger` had an `operator<<`. It would make this code 1 line shorter :)

tests/cpp/lp_namer/WeightsFileReaderTest.cpp
```diff
@@ -22,7 +22,7 @@ TEST_F(WeightsFileReaderTest, FailsIfWeightsFileDoesNotExist) {
   try {
     weights_file_reader.CheckWeightsFile();
   } catch (const WeightsFileReader::WeightsFileOpenError& e) {
-    EXPECT_EQ(e.what(), expectedErrorString.str());
+    EXPECT_EQ(e.ErrorMessage().c_str(), expectedErrorString.str());
```
flomnes, line 25: It seems like you don't need `c_str()` here, and below.

flomnes (commented):
While I agree that displaying the location of an error is a nice feature, it is my opinion that such details should be handled by the logger itself.

Example
```cpp
Logger::ErrorWithLocation << "blah";
```
Resulting in
> "[Error] src/cpp/lp_namer/source.cpp:332: Error blah"

I'm not sure there would be a way to achieve this easily.

There are at least 2 advantages to this approach

- For the next developments, you will be less likely to forget providing the location the logs
- You don't need to think about it, the logging interface is easier to use without those details

src/cpp/xpansion_interfaces/LogUtils.h
a-zakir: thanks

tbittar (approved):

PR #642, opened 2023-03-27
fix expansion bug

### LINKED ISSUE #641: benders exe fails if expansion doesn't exist
launching benders exe without a pre-existing "expansion" folder leads to an error.
`Impossible d'ouvrir le fichier json "./expansion/out.json"`
We need to check if the 'expansion' folder exists and if not create it.

a-zakir:
fix: create expansion dir if it is not exist


a-zakir:
close #641

tbittar (approved):

PR #648, opened 2023-04-04
read variables from mps

a-zakir:
add ability to read mps with named variables in Problem Generation


tests/end_to_end/lpnamer/test_lpnamerEndToEnd.py
```diff
@@ -83,7 +83,7 @@ def launch_and_compare_lp_with_reference(install_dir, master_mode, test_dir):
     study_dir = test_dir.resolve()
     os.chdir(test_dir.parent)
     launch_command = [str(lp_namer_exe), "-o", str(study_dir), "-a", str(zip_path),
-                      "-e", "contraintes.txt", "-f", master_mode]
+                      "-e", "contraintes.txt", "-f", master_mode, "--unnamed-problems"]
```
JasonMarechal25, line 86: If I understand properly, "unnamed-problems" is "legacy" behaviour ?

src/python/antares_xpansion/antares_driver.py
```diff
@@ -95,8 +99,10 @@ def _set_simulation_name(self):
             self.simulation_name = list_of_dirs[-1]
 
     def _get_antares_cmd(self):
-        return [str(self.antares_exe_path), self.data_dir, self.ANTARES_N_CPU_OPTION, str(self.antares_n_cpu),
-                self.zip_option]
+        cmd = [str(self.antares_exe_path), self.data_dir, self.ANTARES_N_CPU_OPTION, str(self.antares_n_cpu), self.zip_option]
+        if(version.parse(__antares_simulator_version__) >= version.parse(self.FIRST_VERSION_WITH_NAMED_PROBLEMS)):
+            cmd.append("--named-problems")
```
JasonMarechal25: For legacy studies, keep behaviour as it is.
For newer studies, use named-problems.
Is that it ?

Do we always want named problems though? Or do we just want to support the feature?

tests/end_to_end/lpnamer/test_lpnamerEndToEnd.py
a-zakir: exactly, reference mps in theses tests are unnamed

src/python/antares_xpansion/antares_driver.py
a-zakir: well, for simulator v8.7 and newer, we're forced to use named mps because variables.txt files are no longer produced.

JasonMarechal25 (approved):

PR #666, opened 2023-07-20
accept study as plain dir or zip

### LINKED ISSUE #665: Look for the study output both in folder and zip file for sensitivity analysis
When launching sensitivity analysis with an output `<my-output>`, Xpansion looks for a zip file `<my-output>.zip`. In some workflows, the output may be decompressed, an the zip file may not exist. This causes Xpansion to crash as it does not find the output.

However, in such case, the folder `<my-output>` exists, therefore we also expect Xpansion to look for a normal folder in addition to the zip file. We may even look for the folder before the zip as if both exists, we should avoid decompressing the archive and work directly on the folder.

a-zakir:
close #665

### DIFF
--- a/src/python/antares_xpansion/config_loader.py
+++ b/src/python/antares_xpansion/config_loader.py
@@ -84,22 +84,14 @@ class ConfigLoader:
             self._xpansion_simulation_name = self._config.simulation_name
 
         else:
-            tmp_zip = Path(self.antares_output()) / \
+            tmp_path = Path(self.antares_output()) / \
                 self._config.simulation_name
-            if self.is_antares_study(tmp_zip):
-                self._last_study = tmp_zip
-                if self.step() in ["resume", "sensitivity"] :
-                    self._xpansion_simulation_name = self._last_study
-                    if self.is_zip(self._last_study):
-                        self._xpansion_simulation_name = self._last_study.parent / self._last_study.stem
-                        with zipfile.ZipFile(self._last_study, 'r') as output_zip:
-                            output_zip.extractall(self._xpansion_simulation_name)
-                else:
-                    self._xpansion_simulation_name = self._last_study.parent / \
-                        (self._last_study.stem+"-Xpansion")
+            if self.is_antares_study_output(tmp_path):
+                self._last_study = tmp_path
+                self._set_xpansion_simulation_name()
             else:
                 raise ConfigLoader.InvalidSimulationName(
-                    f"{tmp_zip} is not a valid zip archive")
+                    f"{tmp_path} is not a valid Antares output")
 
     def _restore_launcher_options(self):
         with open(self.launcher_options_file_path(), "r") as launcher_options:
@@ -511,8 +503,10 @@ class ConfigLoader:
         """
         self._last_study = self.last_modified_study(self.antares_output())
 
-        if self.step() in ["resume", "sensitivity"] :
-            
+        self._set_xpansion_simulation_name()
+
+    def _set_xpansion_simulation_name(self):
+        if self.step() in ["resume", "sensitivity"] : 
             self._xpansion_simulation_name = self._last_study
             if self.is_zip(self._last_study):
                 self._xpansion_simulation_name = self._last_study.parent / self._last_study.stem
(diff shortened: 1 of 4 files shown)


src/python/antares_xpansion/config_loader.py
```diff
@@ -79,15 +81,22 @@ def _set_simulation_name(self):
             raise ConfigLoader.MissingSimulationName(
                 "Missing argument simulationName")
         elif self._config.simulation_name == "last":
-            self._simulation_name = self._config.simulation_name
+            self._xpansion_simulation_name = self._config.simulation_name
 
         else:
             tmp_zip = Path(self.antares_output()) / \
```
tbittar: Rename `tmp_zip` to `tmp` as this no longer necessarily a zip ?

src/python/antares_xpansion/config_loader.py
```diff
@@ -79,15 +81,22 @@ def _set_simulation_name(self):
...
+                    if self.is_zip(self._last_study):
+                        self._xpansion_simulation_name = self._last_study.parent / self._last_study.stem
+                        with zipfile.ZipFile(self._last_study, 'r') as output_zip:
+                            output_zip.extractall(self._xpansion_simulation_name)
+                else:
+                    self._xpansion_simulation_name = self._last_study.parent / \
+                        (self._last_study.stem+"-Xpansion")
             else:
                 raise ConfigLoader.InvalidSimulationName(
                     f"{tmp_zip} is not a valid zip archive")
```
tbittar: The error message may be misleading as `tmp_zip` is not necessarily a zip archive. Maybe something like `tmp is not a valid Antares output` would be more appropriate.

src/python/antares_xpansion/config_loader.py
```diff
@@ -508,44 +509,47 @@ def _set_last_simulation_name(self):
...
 
-        if self.step() == "resume":
-            self._simulation_name = last_dir_path
+        if self.step() in ["resume", "sensitivity"] :
+            
+            self._xpansion_simulation_name = self._last_study
+            if self.is_zip(self._last_study):
+                self._xpansion_simulation_name = self._last_study.parent / self._last_study.stem
+                with zipfile.ZipFile(self._last_study, 'r') as output_zip:
+                    output_zip.extractall(self._xpansion_simulation_name)
```
tbittar, line 514: Lines 514 - 523 seem to be a duplicate of lines 91 - 99, is it possible to factorize ?

src/python/antares_xpansion/config_loader.py
```diff
@@ -79,15 +81,22 @@ def _set_simulation_name(self):
...
         else:
             tmp_zip = Path(self.antares_output()) / \
                 self._config.simulation_name
-            if self.is_zip(tmp_zip):
-                self._last_zip = tmp_zip
-                self._simulation_name = self._last_zip.parent / \
-                    (self._last_zip.stem+"-Xpansion")
+            if self.is_antares_study(tmp_zip):
+                self._last_study = tmp_zip
+                if self.step() in ["resume", "sensitivity"] :
```
tbittar: Lines 91 - 99 look very much like the code of `_set_options_for_benders_solver`, see comment below

src/python/antares_xpansion/config_loader.py
```diff
@@ -508,44 +509,47 @@ def _set_last_simulation_name(self):
...
+        return ext == ".zip" 
+    
+    def update_last_study_with_sensitivity_results(self):
+        if self.is_zip(self._last_study):
+            os.remove(self._last_study)
+            shutil.make_archive(self._last_study.parent / self._last_study.stem, 'zip', self._xpansion_simulation_name)  
+            if(os.path.exists(self._xpansion_simulation_name)):
+                shutil.rmtree(self._xpansion_simulation_name)
+
+    def is_antares_study(self, study):
```
tbittar: We are looking for output folders / archive here ? If so, `is_antares_study_output` may be a better name

src/python/antares_xpansion/config_loader.py
```diff
@@ -508,44 +509,47 @@ def _set_last_simulation_name(self):
...
+            list_of_studies,
             key=lambda x: os.path.getmtime(
                 os.path.join(root_dir, x)),
         )
-
-        last_zip = Path(root_dir) / zip_sorted[-1]
-        filename, ext = os.path.splitext(last_zip)
-        output_dir = os.path.basename(filename)
-        return output_dir
+        last_study = Path(root_dir) / sort_studies[-1]
```
tbittar, line 548: We often have unclear error messages here, like `index out of range` if the list `sort_studies` is empty. Is it possible to add a more clear message like `No Antares output has been found` or something equivalent ?

src/python/antares_xpansion/sensitivity_driver.py
```diff
@@ -36,8 +36,9 @@ def launch(
         self.simulation_output_path = self._get_simulation_output_path(
             simulation_output_path
         )
-        with zipfile.ZipFile(self.simulation_output_path, 'r') as output_zip:
-            output_zip.extractall(xpansion_simulation_output)
+        # if (os.path.splitext(simulation_output_path)[1] == ".zip"):
+        #     with zipfile.ZipFile(self.simulation_output_path, 'r') as output_zip:
+        #         output_zip.extractall(xpansion_simulation_output)
```
tbittar: Remove this code ?

tbittar (changes requested):

src/python/antares_xpansion/sensitivity_driver.py
a-zakir: done before your comment but it was not visbible in github!

tbittar (approved):

PR #669, opened 2023-08-01
improve json outputs

a-zakir:
add cumulative number of sub problem resolved at each iteration

### DIFF
--- a/src/cpp/xpansion_interfaces/ILogger.h
+++ b/src/cpp/xpansion_interfaces/ILogger.h
@@ -63,22 +63,6 @@ struct LogData {
   double master_time;
   double subproblem_time;
   int cumulative_number_of_subproblem_resolved;
-  bool operator==(const LogData &lhs) const {
-    return lb == lhs.lb && best_ub == lhs.best_ub && ub == lhs.ub &&
-           it == lhs.it && best_it == lhs.best_it &&
-           subproblem_cost == lhs.subproblem_cost &&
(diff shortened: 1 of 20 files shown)


src/cpp/xpansion_interfaces/ILogger.h
```diff
@@ -74,7 +75,9 @@ struct LogData {
            max_iterations == lhs.max_iterations &&
            benders_elapsed_time == lhs.benders_elapsed_time &&
            master_time == lhs.master_time &&
-           subproblem_time == lhs.subproblem_time;
+           subproblem_time == lhs.subproblem_time &&
```
tbittar: Here we do a lot of equality comparison between float values which is very risky, we should replace equality between floats by something like abs(lhs - rhs) < epsilon. Do we compare `LogData` only in tests or elsewhere ?

tbittar (commented):
- When we do not use `benders_by_batch` but `benders`, `_data.cumulative_number_subproblems_solved` is uninitialized which causes weird outputs in the json file. We should also compute this data in the classical benders case (at each iteration it is simply increased by the number of subproblems solved)
- We should also complete tests in `json_writer_test.cpp` and `logger_test.cpp` to handle cumulative number of subproblems where needed

src/cpp/xpansion_interfaces/ILogger.h
a-zakir: yes it's only used for test, i've move it to the test src

JasonMarechal25 (approved):

PR #691, opened 2023-10-04
Add documentation on how to publish release

JasonMarechal25:


JasonMarechal25:
PDF is not the best way for our usage. Looking into another solution

PR #710, opened 2023-10-24
Create lp directories in cpp

JasonMarechal25:
- Don't delete old log
- Don't delete existing -xpansion directory
- Create -xpansion and lp directories during problem generation CPP
- Remove "dead" code path using results from file or from antares: simplify, only keep zip path

### DIFF
--- a/src/python/antares_xpansion/problem_generator_driver.py
+++ b/src/python/antares_xpansion/problem_generator_driver.py
@@ -3,6 +3,7 @@
 """
 
 import os
+import shutil
 import subprocess
 import sys
 from dataclasses import dataclass
@@ -65,6 +66,7 @@ class ProblemGeneratorDriver:
         """
             problem generation step : getnames + lp_namer
         """
+        self.clear_old_log()
         self.logger.info("Problem Generation")
         self.output_path = output_path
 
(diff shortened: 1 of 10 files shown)


src/python/antares_xpansion/problem_generator_driver.py
```diff
@@ -80,9 +78,7 @@ def set_output_path(self, output_path):
         if output_path.exists():
             self._output_path = output_path
             self.xpansion_output_dir = output_path.parent / \
-                (output_path.stem+"-Xpansion")
-            if self.xpansion_output_dir.exists():
-                shutil.rmtree(self.xpansion_output_dir)
+                                       (output_path.stem + "-Xpansion")
```
a-zakir, line 83: This Pr is about creating -xpansion and lp dir so why you modify the python code?
else we face in some case memory issues and i think it's a good practice to clean the xpansion dir before the problem_generation run.

src/python/antares_xpansion/problem_generator_driver.py
```diff
@@ -112,8 +104,6 @@ def _lp_step(self):
                 "ERROR: exited lpnamer with status %d" % returned_l.returncode)
 
     def create_lp_dir(self):
-        if os.path.isdir(self._lp_path):
```
a-zakir: i suggest to keep this code

a-zakir (changes requested):

a-zakir (approved):

PR #711, opened 2023-10-25
deploy doxygen code documentation

a-zakir:


tbittar (commented):
Two questions :
- What is the purpose of the `.nojekyll` file ?
- Would it be interesting to add a link from the readthedocs documentation to the classes page of the doxygen docs ?

a-zakir:
> Two questions :
> 
>     * What is the purpose of the `.nojekyll` file ?
> 
>     * Would it be interesting to add a link from the readthedocs documentation to the classes page of the doxygen docs ?

1- to bypass jekyll, ,since it's ran automatically by gh-pages
2- yes let's do that

tbittar (approved):

PR #712, opened 2023-10-30
Simplify Problem generation commande line API

### LINKED ISSUE #703: Have simpler/less parameters for ProblemGeneration

JasonMarechal25:


a-zakir:
i am fully ok with the idea, but i think it would be simpler to have all PbGen options in a file (*.json) as Benders exe?

PR #734, opened 2023-12-15
fix xpress custom install

a-zakir:


src/cpp/multisolver_interface/environment.cc
```diff
@@ -336,7 +336,7 @@ bool initXpressEnv(bool verbose, int xpress_oem_license_key) {
       return false;
     }
   } else {
-    xpresspath = xpress_from_env;
+    xpresspath = (std::filesystem::path(xpress_from_env) / "bin").string();
```
tbittar: The Xpress documentation advises to call `XPRSinit` (line 351) with a `NULL` argument instead of `xpresspath`. And if we have to use `xpresspath` it should be preferable to use the environment variable `XPAUTH_PATH`. As `XPAUTH_PATH` gives the full path of the license file, we should just get the directory from the full path.

tbittar (changes requested):

src/cpp/multisolver_interface/environment.cc
a-zakir: the XPRSinit is called with the null argument when the solver is really called(see SolverXpress.cpp)
this function searches Xpress lib :
- in "standard" place like /opt/xpressmp 
- by looking in the env variable XPRESSDIR
the whole thing is to not be tied with an xpress version so no need to set the XPAUTH_PATH from xpvars.sh script.
there is a scenario where the license and binaries are in different places, how would you get the lib, bin from license file?

JasonMarechal25:
No change left. Changes were integrated in another PR

PR #736, opened 2023-12-19
C4 architecture diagrams

JasonMarechal25:
For architecture diagrams:

- System Context Diagram : provide a high level view of the system surrounding Antares
- Container Diagram : provide context on Antares platform components
- Container Diagram (simple) : same as above, removing details for clarity
- Component Diagram : provide a zoomed view of Xpansion components architecture


sylvlecl (approved):
Nice work, I don't have much to say about the content.
This documentation could be merged as is without any problem. The diagrams except the xpansion ones are specific to RTE deployment and usage though (windows VM etc), we could re-use them in a more internal documentation.

PR #738, opened 2023-12-26
merge classical and by batch benders

### LINKED ISSUE #727: merge classical benders with by batch
Goal:
one benders exe, the choice of the algorithm have to be determined by the batch size
@tbittar

a-zakir:
Goal:
one benders executable instead the classical benders and the by batch one.
how:
the choice of the algorithm is decided by the batch_size parameter

close #727


src/cpp/benders/factories/BendersFactory.cpp
```diff
@@ -35,7 +34,11 @@ int RunBenders(char** argv, const std::filesystem::path& options_file,
         std::filesystem::path(options.OUTPUTROOT) / "reportbenders.txt";
 
     Writer writer;
-
+    const auto couplig_map = build_input(benders_options.STRUCTURE_FILE);
+    const auto method = (options.BATCH_SIZE == 0 ||
```
JasonMarechal25: Can you extract in a function with if for readability and debugging purpose please

src/cpp/benders/factories/BendersFactory.cpp
```diff
@@ -49,18 +52,16 @@ int RunBenders(char** argv, const std::filesystem::path& options_file,
       writer = build_void_writer();
     }
 
+    world.barrier();
     pBendersBase benders;
     if (method == BENDERSMETHOD::BENDERS) {
       benders = std::make_shared<BendersMpi>(benders_options, logger, writer,
                                              env, world);
-    } else if (method == BENDERSMETHOD::BENDERSBYBATCH) {
+    } else {
```
JasonMarechal25: Since BENDERSMETHOD is an enum (good), it would be best to employ a switch case and keep the "else" clause in a default case.
It's "just in case" a new value need implementation

JasonMarechal25 (changes requested):
Good
Just a couple of nitpicking comments

JasonMarechal25:
Don't forget to add some details Why/What/How in the top comment

src/cpp/benders/factories/BendersFactory.cpp
tbittar: And also typo `couplig_map` -> `coupling_map`

src/cpp/benders/factories/BendersFactory.cpp
tbittar: By the way, the enum also has a possible value `MERGEMPS` (but I do not know if `BENDERSMETHOD` can even take this value ?)

When using merge_mps, we use the merge_mps exe, which uses `MergeMPS.cpp` 

So maybe this enum value is not relevant ? We need to think about it

src/cpp/benders/factories/BendersFactory.cpp
```diff
@@ -35,7 +34,11 @@ int RunBenders(char** argv, const std::filesystem::path& options_file,
         std::filesystem::path(options.OUTPUTROOT) / "reportbenders.txt";
 
     Writer writer;
-
+    const auto couplig_map = build_input(benders_options.STRUCTURE_FILE);
+    const auto method = (options.BATCH_SIZE == 0 ||
+                         options.BATCH_SIZE == couplig_map.size() - 1)
```
tbittar: What should we do if the user sets a value `batch_size > nb_subproblems`, Benders or Benders by batch ? Still need to think about it

src/cpp/benders/factories/BendersFactory.cpp
a-zakir: two things:
the "method" option in CLI (xpansion-launcher) expect either (benders or mergemps) while the fullrun or the benders factory set only BENDERSMETHOD to {BENDERS or *BYBATCH}
i will clarify usage

src/cpp/benders/factories/BendersFactory.cpp
a-zakir: it's bendersbybatch and end up with an error

src/cpp/benders/factories/BendersFactory.cpp
a-zakir: used switch case  && deleted MERGE_MPS

JasonMarechal25 (approved):

src/cpp/benders/factories/BendersFactory.cpp
```diff
@@ -12,6 +11,14 @@
 #include "gflags/gflags.h"
 #include "glog/logging.h"
 
+BENDERSMETHOD DeduceBenderMethod(size_t coupling_map_size, size_t batch_size) {
```
tbittar: `Benders` is the name of a mathematician, do not forget the final `s`

tests/CMakeLists.txt
```diff
@@ -12,9 +12,9 @@ find_python_module(pytest)
 find_python_module(numpy)
 
 if (PYTHON_MODULE_pytest_FOUND AND PYTHON_MODULE_numpy_FOUND)
-        set(xpress_avalaible "False")
+        set(xpress_avalaible "")
         if(${XPRESS})
-	        set(xpress_avalaible "True")
+	        set(xpress_avalaible "--xpress")
```
tbittar: `available` instead of `avalaible`

tbittar (changes requested):
Just 2 typos

src/cpp/benders/factories/BendersFactory.cpp
a-zakir: thanks

tbittar (approved):

Issue #742, opened 2024-01-05
remove Xpress OFF ci


a-zakir:
we have for each ci Xpress Matrix [on/off] since Xpansion is no longer linking against Xpress lib we can remove the Xpress OFF matrix.

PR #764, opened 2024-02-29
Produce only one asset (with xpress)

JasonMarechal25:
Given xpress is now dynamically loaded at runtime it is no more necessary to produce several assets.


a-zakir:
close #742

a-zakir (approved):

PR #793, opened 2024-04-26
Update ubuntu badge

JasonMarechal25:


flomnes:
image
That really what we want ?

tbittar (changes requested):
The same changes must be done in `docs/index.md` so that modifications also apply in readthedocs

tbittar (approved):

PR #794, opened 2024-04-26
[API] Cleanup code

JasonMarechal25:
- Add dome no_discard
- Use unsigned where it make sense
- Avoid implicit cast and use static_cast when possible
- Solver factory initialize all solvers (previously missing for xpress)


src/cpp/lpnamer/problem_modifier/AdditionalConstraints.h
```diff
@@ -113,7 +113,7 @@ struct AdditionalConstraints
    * constraints file path
    */
   explicit AdditionalConstraints(
-      std::string const& constraints_file_path,
+      std::string constraints_file_path,
```
tbittar, line 116: Why is it not necessary to have a const& ?

tbittar (commented):
As you perform `solver->init()` in create solver, you may be able to remove all calls to `solver->init()` in Worker.cpp and tests

src/cpp/lpnamer/helper/ColumnToChange.h
```diff
@@ -9,7 +9,7 @@ struct ColumnToChange {
   bool operator==(const ColumnToChange& other) const;
 
   colId id;
-  int time_step;
+  unsigned int time_step;
```
flomnes, line 12: `size_t` or `unsigned` ? See below

src/cpp/lpnamer/problem_modifier/AdditionalConstraints.h
flomnes: Because he `std::moves` a copy
1. Pass by copy
2. `std::move` the copy to our data member

A concurrent approach is (usual approch)
1. Pass by const ref
2. Copy

They're equivalent I AFAIK

src/cpp/multisolver_interface/SolverFactory.cpp
```diff
@@ -69,41 +71,47 @@ SolverAbstract::Ptr SolverFactory::create_solver(
...
   } else if (solver_name == CBC_STR) {
-    return std::make_shared<SolverCbc>(log_manager);
+    ret = std::make_shared<SolverCbc>(log_manager);
   }
 #endif
   else {
     throw InvalidSolverNameException(solver_name, LOGLOCATION);
   }
+  ret->init();
+  return ret;
```
flomnes, line 114: Slight change of behavior here. `init` wasn't called at all before, but now we call it. Did you forget to remove a few calls at higher levels ?

src/cpp/lpnamer/helper/ColumnToChange.h
JasonMarechal25: unsigned

src/cpp/multisolver_interface/SolverFactory.cpp
JasonMarechal25: I've now removed all call to init at higher level (after calls to factory) reducing temporal coupling

tbittar (approved):

PR #795, opened 2024-04-26
[API] Refacto test LPNamer

JasonMarechal25:
- Add copy of SmallTestFiveCandidates in data tests
- Improve code readability
- Use parametrize to create test combination instead of doing it by hand
- Work in temp dir instead of source dir


data_test/tests_lpnamer/SmallTestFiveCandidates/reportbenders.txt
tbittar, line 1: Do we need this file ?

tbittar (approved):

tbittar (approved):

data_test/tests_lpnamer/SmallTestFiveCandidates/reportbenders.txt
JasonMarechal25: I don't think so. Removed

PR #753, opened 2024-01-25
Feature/antares in memory

JasonMarechal25:
Introduce "study mode". A way to run xpansion and problem generation with a study folder as parameter. In this case Antares will not be run as a stand-alone but as a library by problem generation.

- New ADRs
- Add a "new" data set for lpnamer E2E tests: SmallTestFiveCandidates, a copy of the example provided by Xpansion.
- Refactor test_lpnamerEndToEnd.py : when possible prefer the use multiple @parametrize instead of writing the combinations in a list. 
- Refactor test_lpnamerEndToEnd.py : copy the test data in a temporary folder and works there. Prevent creating outputs in source code.
- Add a test case for Study mode in test_lpnamerEndToEnd.py
- Add a new E2E test: short memory
- Refactor ProblemGenerationExeOptionsTest to test several combinations of mutually exclusive parameters
- In SolverFactory.cpp call Init() on solver objects before returning them. Prevent temporal coupling for Xpress solver where a call to init() was mandatory.
- Add Antares as a build dependency

Missing pieces (every thing is open to discussion):

- [ ] Handle Antares error #888 
- [x] Update general data ini
- [ ] Revert general data ini #889
- [x] Write ADR
- [ ] Update sequence diagram #890
- [ ] Maybe update C4 #890
- [x] Use enum mode in LPFileExtractor #797 
- [ ] find/define a way to log before lp dir exist (for antares error for exemple) #888 
- [ ] Handle variables.txt directly at Lps level (SignificationMetierDesVariables)
- [ ] Factorise variables name and "calculate" time step with week number on xpansion side (A faire dans Antares bien et vite)
- [x] Cleanup python to only support memory mode (See ADR, we keep some support for now)
- [x] antares named problèmes force mps writing. Peut etre pas besoin de l'option named sachant que les variables sont nommées dans la structure LP
- [ ] Mettre à jour les docs: notamment comment build, dépendances, etc #893 
- [ ] Investigate to use full run executable with --study option and use fullrun driver #892 
- [x] Handle PBGen step with memory
- [ ] Add entry in multisolver to rename a range of col/var https://github.com/AntaresSimulatorTeam/antares-xpansion/pull/753#discussion_r1494876129 #891
- [ ] Run Antares in parallele mode

Following other PR:

- [ ] Don't overly rename problem in API mode

PR #797, opened 2024-04-26
[API] Better handling of execution modes

JasonMarechal25:
- Properly distinguish between archive, file and study mode
- study mode (using antares lib) not implemented yet
- Code cleanup
- Refactor how to handle mandatory but exclusive parameters. Hopefully make it easier to add new one in the future


src/cpp/lpnamer/main/ProblemGeneration.cpp
```diff
@@ -44,30 +43,57 @@ void CreateDirectories(const std::filesystem::path& output_path) {
...
 std::filesystem::path ProblemGeneration::updateProblems() {
-  const auto xpansion_output_dir = options_.XpansionOutputDir();
+  using namespace std::string_literals;
+  std::filesystem::path xpansion_output_dir;
   const auto archive_path = options_.ArchivePath();
 
-  auto deduced_xpansion_output_dir =
-      options_.deduceXpansionDirIfEmpty(xpansion_output_dir, archive_path);
+  if (mode_ == Mode::ARCHIVE) {
+    xpansion_output_dir =
```
flomnes: Input mode ? Mode is too vague

tests/cpp/lp_namer/ProblemGenerationExeOptionsTest.cpp
```diff
@@ -65,25 +70,62 @@ TEST_F(ProblemGenerationExeOptionsTest, MasterFormulationDefaultValue) {
...
+template <typename T1, typename T2, typename... Ts>
+auto flattenPairs(const std::pair<T1, T2>& pair, const Ts&... rest) {
+  return std::tuple_cat(std::make_tuple(pair.first, pair.second),
+                        flattenPairs(rest...));
+}
+
+auto cases() {
+  using namespace std::string_literals;
+  return ::testing::Combine(
+      ::testing::Values(std::make_pair("--archive"s, "archive.zip"s),
```
flomnes: `std::make_pair` was useful before C++17's template deduction for classes (deduction only worked for functions). Now we can just use `std::pair`

data_test/tests_lpnamer/SmallTestFiveCandidates/reportbenders.txt
tbittar, line 1: Remove file ?

src/cpp/lpnamer/input_reader/LpFilesExtractor.cpp
```diff
@@ -24,11 +51,20 @@ void LpFilesExtractor::ExtractFiles() const {
...
-                          xpansion_output_dir_ / "area.txt");
 
-  const auto vect_interco_files =
-      archive_reader.ExtractPattern("interco*.txt", "", xpansion_output_dir_);
+  if (mode_ == Mode::ARCHIVE) {
+    std::filesystem::rename(vect_area_files[0],
+                            xpansion_output_dir_ / "area.txt");
+  } else {
+    try {
+      std::filesystem::copy(vect_area_files[0],
```
tbittar: Why a copy rather than just rename ?

src/cpp/lpnamer/main/ProblemGeneration.cpp
```diff
@@ -44,30 +43,57 @@ void CreateDirectories(const std::filesystem::path& output_path) {
...
+    throw std::runtime_error("Mode API Not implemented yet");
+  }
+
+  if (mode_ == Mode::FILE) {
+    simulation_dir_ = options_.XpansionOutputDir();  // Legacy naming.
+    // options_.XpansionOutputDir() point in fact to a simulation output from
+    // antares
+  }
+
+  if (mode_ == Mode::ANTARES_API || mode_ == Mode::FILE) {
```
tbittar: Why `Mode::ANTARES_API` here as we would have already raised the not implemented error ? For the future maybe ?

src/cpp/lpnamer/main/ProblemGeneration.cpp
```diff
@@ -44,30 +43,57 @@ void CreateDirectories(const std::filesystem::path& output_path) {
...
+    xpansion_output_dir = simulation_dir_;
+  }
 
   const auto log_file_path =
-      deduced_xpansion_output_dir / "lp" / "ProblemGenerationLog.txt";
+      xpansion_output_dir / "lp"s / "ProblemGenerationLog.txt"s;
 
-  CreateDirectories(deduced_xpansion_output_dir);
+  CreateDirectories(xpansion_output_dir);  // Ca ou -Xpansion ?
```
tbittar, line 83: Does the Python part (or any other part of the code) rely on the `-Xpansion` suffix ?

src/cpp/lpnamer/input_reader/LpFilesExtractor.cpp
```diff
@@ -1,14 +1,41 @@
...
+    vect_area_files =
+        archive_reader.ExtractPattern("area*.txt", "", xpansion_output_dir_);
+
+    vect_interco_files =
+        archive_reader.ExtractPattern("interco*.txt", "", xpansion_output_dir_);
+
+    archive_reader.ExtractPattern("ts-numbers*", "", xpansion_output_dir_);
+    archive_reader.Close();
+    archive_reader.Delete();
+  } else {
```
tbittar: Why not be more explicit and putting `else if Mode::ANTARES_API || Mode::FILE`  ? As there is also an unknown mode (but error would have been raised before ?)

src/cpp/lpnamer/main/ProblemGeneration.cpp
```diff
@@ -187,6 +223,8 @@ void ProblemGeneration::RunProblemGeneration(
       antares_version < first_version_without_variables_files;
   (*logger)(LogUtils::LOGLEVEL::INFO)
       << "rename problems: " << std::boolalpha << rename_problems << std::endl;
+
+
   auto files_mapper = FilesMapper(antares_archive_path, xpansion_output_dir);
   auto mpsList = files_mapper.MpsAndVariablesFilesVect();
```
tbittar, line 229: Should we do this in ANTARES_API mode ?

src/cpp/lpnamer/main/ProblemGenerationExeOptions.cpp
```diff
@@ -21,19 +23,33 @@ ProblemGenerationExeOptions::ProblemGenerationExeOptions()
...
+        "Need to give at least on of [OutputDir, Archive, Study] options"s;
     throw ProblemGenerationOptions::MissingParameters(msg, log_location);
   }
 }
+void ProblemGenerationExeOptions::checkOnlyOneMandatoryOption(
+    const std::string& log_location) const {
+  if (std::vector<std::string> args = {this->XpansionOutputDir().string(),
+                                       this->ArchivePath().string(), this->StudyPath().string()};
+      std::ranges::count_if(args, std::ranges::empty) < (args.size() - 1)) {
+    auto msg = "Only one of [archive, output, study] parameters is accepted"s;
```
tbittar, line 46: - Uniformize error message with the one of the previous function
- Would it be possible to have only one function that checks there is only one and exactly one option specified (as you check at least one and no more than 1) ?

src/cpp/lpnamer/main/ProblemGenerationExeOptions.cpp
```diff
@@ -21,19 +23,33 @@ ProblemGenerationExeOptions::ProblemGenerationExeOptions()
...
+    auto msg =
+        "Need to give at least on of [OutputDir, Archive, Study] options"s;
     throw ProblemGenerationOptions::MissingParameters(msg, log_location);
   }
 }
+void ProblemGenerationExeOptions::checkOnlyOneMandatoryOption(
+    const std::string& log_location) const {
+  if (std::vector<std::string> args = {this->XpansionOutputDir().string(),
+                                       this->ArchivePath().string(), this->StudyPath().string()};
+      std::ranges::count_if(args, std::ranges::empty) < (args.size() - 1)) {
```
tbittar: Why `args.size() - 1` and not `2` as it seems that the condition does not really depend on the number of elements in `args` ?

src/cpp/lpnamer/main/ProblemGenerationExeOptions.cpp
```diff
@@ -42,13 +58,13 @@ std::filesystem::path ProblemGenerationExeOptions::deduceArchivePathIfEmpty(
     if (xpansion_output_dir.string().find("-Xpansion") == std::string::npos) {
       auto log_location = LOGLOCATION;
       auto msg =
-          "Archive path is missing and output path does not contains"
-          " \"-Xpansion\" suffixe. Can't deduce archive file name.";
+          "Archive path is missing and output path does not contains"s
+          " \"-Xpansion\" suffixe. Can't deduce archive file name."s;
```
tbittar: suffix

src/cpp/lpnamer/main/ProblemGenerationExeOptions.cpp
```diff
@@ -21,19 +23,33 @@ ProblemGenerationExeOptions::ProblemGenerationExeOptions()
...
+    const std::string& log_location) const {
+  if (std::vector<std::string> args = {this->XpansionOutputDir().string(),
+                                       this->ArchivePath().string(), this->StudyPath().string()};
+      std::ranges::count_if(args, std::ranges::empty) < (args.size() - 1)) {
+    auto msg = "Only one of [archive, output, study] parameters is accepted"s;
+    throw ProblemGenerationOptions::ConflictingParameters(msg, log_location);
+  }
+}
 
 std::filesystem::path ProblemGenerationExeOptions::deduceArchivePathIfEmpty(
```
tbittar: This function seems unused

tbittar (changes requested):

src/cpp/lpnamer/main/ProblemGenerationExeOptions.cpp
flomnes: Indeed, looking at the code, `args.size()` is always 3 regardless of it's content. I agree that writing 2 instead of `args.size() - 1` would make the code more clear.

src/cpp/lpnamer/main/ProblemGenerationExeOptions.cpp
```diff
@@ -31,9 +31,9 @@ void ProblemGenerationExeOptions::Parse(unsigned int argc,
 }
 void ProblemGenerationExeOptions::checkAtLeastOneMandatoryOption(
     const std::string& log_location) const {
-  if (std::vector<std::string> mandatory = {this->XpansionOutputDir(),
-                                            this->ArchivePath(),
-                                            this->StudyPath()};
+  if (std::vector<std::string> mandatory = {this->XpansionOutputDir().string(),
+                                            this->ArchivePath().string(),
+                                            this->StudyPath().string()};
```
flomnes: Code duplication. Create a small static/lambda function to build this triplet ?

src/cpp/lpnamer/main/ProblemGeneration.cpp
JasonMarechal25: yes

src/cpp/lpnamer/input_reader/LpFilesExtractor.cpp
JasonMarechal25: Indeed.

Now properly supported and future prof with default case

src/cpp/lpnamer/main/ProblemGenerationExeOptions.cpp
JasonMarechal25: The rational was that if a new parameter is introduced, ards.size()-1 becomes 3 so instead of making 2 changes (list of args and size of args) using args.size()-1 made it that only changing the list was necessary.
But it's now reworled

src/cpp/lpnamer/main/ProblemGenerationExeOptions.cpp
JasonMarechal25: Code was simplify according to suggestion

src/cpp/lpnamer/input_reader/LpFilesExtractor.cpp
```diff
@@ -6,93 +6,195 @@
...
-            if (entry.path().filename().string().starts_with("interco")) {
-              vect_interco_files.push_back(entry.path());
+/**
+ * @brief This method is used to get the files from the simulation directory.
+ *
+ * The patterns it looks for are "area*.txt" and "interco*.txt".
+ *
+ * @return A pair of vectors containing the paths of the files. The first vector contains the paths of the "area*.txt" files and the second vector contains the paths of the "interco*.txt" files.
+ */
+LpFilesExtractor::areaAndIntecoPaths LpFilesExtractor::getFiles() const{
```
tbittar: IntercoPaths

src/cpp/lpnamer/input_reader/LpFilesExtractor.cpp
```diff
@@ -6,93 +6,195 @@
...
+}
+
+/**
+ * @brief This method is used to extract files from an archive.
+ *
+ * The patterns it looks for are "area*.txt", "interco*.txt", and "ts-numbers*". The extracted files are stored in the `xpansion_output_dir_` directory.
+ *
+ * @return A pair of vectors containing the paths of the extracted files. The first vector contains the paths of the "area*.txt" files and the second vector contains the paths of the "interco*.txt" files.
+ */
+LpFilesExtractor::areaAndIntecoPaths LpFilesExtractor::getFilesFromArchive() const {
```
tbittar: Interco

src/cpp/lpnamer/input_reader/LpFilesExtractor.cpp
```diff
@@ -6,93 +6,195 @@
...
+  auto archive_reader = ArchiveReader(this->antares_archive_path_);
+  archive_reader.Open();
+  vect_area_files =
+      archive_reader.ExtractPattern("area*.txt", "", this->xpansion_output_dir_);
+
+  vect_interco_files = archive_reader.ExtractPattern("interco*.txt", "",
+                                                     this->xpansion_output_dir_);
+
+  archive_reader.ExtractPattern("ts-numbers*", "", this->xpansion_output_dir_);
+  return std::make_pair(vect_area_files, vect_interco_files);
```
tbittar: Don't we need to close and delete archive ?

src/cpp/lpnamer/input_reader/LpFilesExtractor.cpp
```diff
@@ -6,93 +6,195 @@
...
+/**
+ * @brief This method checks if the number of interco files is correct.
+ *
+ * We should have only one file named "interco*.txt".
+ *
+ * @param vect_interco_files
+ */
+void LpFilesExtractor::checkProperNumberOfIntercoFiles(
+    const std::vector<std::filesystem::path>& vect_interco_files) const {
   if (auto num_intercos_file = vect_interco_files.size();
```
tbittar: `num_interco_file` is more coherent

src/cpp/lpnamer/input_reader/LpFilesExtractor.h
```diff
@@ -35,5 +35,16 @@ class LpFilesExtractor {
   class ErrorWithIntercosFile : public LogUtils::XpansionError<std::runtime_error> {
     using LogUtils::XpansionError<std::runtime_error>::XpansionError;
   };
+ private:
+  using areaAndIntecoPaths = std::pair<std::vector<std::filesystem::path>, std::vector<std::filesystem::path>>;
```
tbittar: Interco

src/cpp/lpnamer/input_reader/LpFilesExtractor.cpp
tbittar, line 1: If we expect only one area and interco file, why do we use a vector to store them and check that length is 1 ? Wouldn't it be simpler to define a single path and raise an error if we try to redefine it or something like that ? This was already  like this before your PR but I think there could improvements here in the future (not to be treated in this PR though)

tbittar (approved):

src/cpp/lpnamer/input_reader/LpFilesExtractor.cpp
JasonMarechal25: archive_reader follow RAII and close/delete when getting destroyed

src/cpp/lpnamer/input_reader/LpFilesExtractor.cpp
JasonMarechal25: Yes it would be good to look into into it.

JasonMarechal25: #815

src/cpp/lpnamer/input_reader/LpFilesExtractor.cpp
JasonMarechal25: During implementation I had some issue with juste rename but it doesn't seem to cause issue now.

src/cpp/lpnamer/main/ProblemGeneration.cpp
JasonMarechal25: Renaming? Probably not. Adding a todo in #753

src/cpp/lpnamer/main/ProblemGeneration.cpp
JasonMarechal25: Yes, but the conditions are foggy

src/cpp/lpnamer/main/ProblemGeneration.cpp
tbittar: No I meant creating the `mpsList` object (it seems unused in ANTARES_API mode) ?

JasonMarechal25: Yes in  the final implementation they're not used for API mode.

PR #817, opened 2024-05-17
Fix criterion computation

tbittar:
The weight of the criterion is only given by `subproblem_weight` which already takes into account the number of scenarios. This PR is just a quick fix, further work will be needed to clean the useless information about scenario retrieval from configuration file, see #818 .


JasonMarechal25 (approved):

a-zakir:
Weigth and number of years are not necessraly the same!

PR #851, opened 2024-06-19
Remove TBB dependency

JasonMarechal25:
With gcc10, TBB isn't required anymore to support std::execution_policies


.github/workflows/ubuntu-system-deps-build.yml
```diff
@@ -39,7 +39,7 @@ jobs:
       - name: Install mandatory system libraries
         run: |
           sudo apt-get update --fix-missing
-          sudo apt-get install libtbb-dev
+          sudo apt-get install
```
a-zakir:
```suggestion
          sudo apt-get install g++-10 gcc-10
```

a-zakir (approved):

PR #854, opened 2024-06-20
Fix vcpkg caching issues

JasonMarechal25:
VCPKG_BINARY_SOURCES env variable was missing in some case. This variable define the path where vcpkg cache things.
Since the variable was not set it used the default path which was different from the path set for cache action.


a-zakir (approved):

a-zakir:
why you didn't merge this pr in develop? @JasonMarechal25

a-zakir:
i mean it'll help

PR #879, opened 2024-07-19
[ANT-1890] Properly handle playlist_reset in generaldata.ini

### LINKED ISSUE #864: [1.3.0] stoi error parsing playlist
In GeneralDataReader.h:GetKeyValFromLine
We parse the section "[playlist]" of generalData.ini

If most values are in the form of "key += Number" one valid entry is "playlist_reset = False"
We use "std::stoi" to convert numbers, however "False" is not a valid representation of a number.

Historically we used std::atoi which does not throw and return 0 in case of error.

std::atoi is useful for playlist_reset entry, but if we have an invalid "key += Number" entry it will silently fail and wrongly set 0 as value for the key.

I think we should handle the exception properly in case Number is invalid and generate meaningful error.
We could however ignore "playlist_reset" line entirely either inside GetKeyValFromLine and return {"", 0} or not call GetKeyValFromLine for playlist_reset line.

Fix should be applied to 1.3.0 release. Check 1.2.x version if bug is present too.

JasonMarechal25:
Fix #864 
Properly handle data and avoid exception thrown when reading not numerical values with stoi when reading "playlist_reset" entry in general_data.ini

### DIFF
--- a/src/cpp/lpnamer/input_reader/GeneralDataReader.cpp
+++ b/src/cpp/lpnamer/input_reader/GeneralDataReader.cpp
@@ -1,8 +1,37 @@
 #include "GeneralDataReader.h"
 
+#include <algorithm>
+#include <filesystem>
+#include <fstream>
 #include <iterator>
+#include <numeric>
+#include <sstream>
+#include <stdexcept>
+#include <vector>
 
+#include "INIReader.h"
 #include "LogUtils.h"
+#include "ProblemGenerationLogger.h"
+#include "StringManip.h"
+class IniReaderUtils {
+ public:
+  static bool LineIsNotASectionHeader(const std::string& line) {
+    return StringManip::split(StringManip::trim(line), '=').size() == 2;
+  }
+
+  static std::string ReadSectionHeader(const std::string& line) {
+    auto str = StringManip::trim(line);
+    str.erase(std::remove(str.begin(), str.end(), '['), str.end());
+    str.erase(std::remove(str.begin(), str.end(), ']'), str.end());
+    return str;
+  }
+
+  static std::pair<std::string, int> GetKeyValFromLine(std::string_view line) {
+    auto key = StringManip::trim(StringManip::split(line, '=')[0]);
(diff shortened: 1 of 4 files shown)


a-zakir (dismissed):

src/cpp/lpnamer/input_reader/GeneralDataReader.cpp
```diff
@@ -1,8 +1,37 @@
...
+  static std::string ReadSectionHeader(const std::string& line) {
+    auto str = StringManip::trim(line);
+    str.erase(std::remove(str.begin(), str.end(), '['), str.end());
+    str.erase(std::remove(str.begin(), str.end(), ']'), str.end());
+    return str;
+  }
+
+  static std::pair<std::string, int> GetKeyValFromLine(std::string_view line) {
+    auto key = StringManip::trim(StringManip::split(line, '=')[0]);
+    auto val = std::stoi(StringManip::trim(StringManip::split(line, '=')[1]));
```
flomnes:
```suggestion
    const auto splitLine = StringManip::split(line, '=');
    auto key = StringManip::trim(splitLine[0]);
    auto val = std::stoi(StringManip::trim(splitLine[1]));
```

tests/cpp/lp_namer/GeneralDataReadetTests.cpp
```diff
@@ -0,0 +1,42 @@
...
+  auto test_dir = std::filesystem::temp_directory_path() / std::tmpnam(nullptr);
+  std::filesystem::create_directory(test_dir);
+  auto ini_file = test_dir / "test.ini";
+  std::ofstream file(ini_file);
+  file << "[general]\n"
+          "nbyears=10\n"
+          "user-playlist=true\n"
+          "[playlist]\n"
+          "playlist_reset = ";
+  GetParam() ? file << "true\n" : file << "false\n";
```
flomnes:
```suggestion
  file << (GetParam() ? "true\n" : "false\n");
```

flomnes (approved):

PR #886, opened 2024-07-31
Test assets are running in fresh environnement (Ubuntu)

JasonMarechal25:
Running tests inside the environment that served to build them and running assets in fresh environment may exhibit different behavior.

With this PR we execute a simple simulation in a fresh environment to ensure they can be run without issue.

### DIFF
--- a/.github/workflows/build_ubuntu.yml
+++ b/.github/workflows/build_ubuntu.yml
@@ -223,6 +223,41 @@ jobs:
           name: user-guide
           path: ${{ steps.create-user-guide.outputs.pdf-path }}
 
+  test_assets:
+    runs-on: ubuntu-latest
+    needs: [ build ]
+    steps:
+      - name: Download build assets
+        uses: actions/download-artifact@v4
+        with:
+          name: ${{needs.build.outputs.TGZ_NAME}}
+      - name: extract
+        run: |
+          tar -xzf ${{needs.build.outputs.TGZ_NAME}}
+      - name: run launcher
+        run: |
+          ./antares-xpansion-launcher -h
+          ./antares-xpansion-launcher -i examples/SmallTestFiveCandidates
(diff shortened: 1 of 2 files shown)


.github/workflows/build_ubuntu.yml
```diff
@@ -229,6 +223,48 @@ jobs:
...
+      - name: extract
+        run: |
+          tar --strip-components=1 -xzf ${{needs.build.outputs.SINGLE_FILE_NAME}}
+      - name: run launcher
+        run: |
+          sudo apt-get update -y
+          sudo apt-get install openmpi-bin -y
+          ls
+          ./antares-xpansion-launcher -h
+          ./antares-xpansion-launcher -i examples/SmallTestFiveCandidates
```
a-zakir, line 270: would be nice to split sequential and parrallel run

.github/workflows/build_ubuntu.yml
```diff
@@ -146,12 +146,6 @@ jobs:
         run: |
           cmake --build _build --config Release -j$(nproc)
 
-      - name: Test
-        run: |
-          export PATH=${GITHUB_WORKSPACE}/_build/vcpkg_installed/x64-linux/tools/openmpi/bin:$PATH
-          cd _build
-          ctest -C Release --output-on-failure -L "medium|unit|benders|lpnamer"
-
```
flomnes: In my opinion, you should run unit tests as well a "full run" tests. Just because "full run" test don't return errors doesn't mean there is no regression.

a-zakir (approved):

.github/workflows/build_ubuntu.yml
a-zakir: can't run tests with the executable,

JasonMarechal25: Yes this is an unwanted modification

PR #896, opened 2024-08-12
Remove sirius port

JasonMarechal25:
Sirius is not a direct dependency, we should not manage it as such. At best/worst we can download it but not manage it


sylvlecl (approved):
Makes sense!

PR #897, opened 2024-08-12
Test assets are running in fresh environnement (Windows)

JasonMarechal25:


.github/workflows/build_ubuntu.yml
```diff
@@ -229,6 +223,49 @@ jobs:
           name: user-guide
           path: ${{ steps.create-user-guide.outputs.pdf-path }}
 
+  test_assets:
+    runs-on: ubuntu-latest
```
a-zakir: duplicated? there is another Pr with the same purpose

JasonMarechal25: Yes it will disapear when merging and updating from develop. Should have set this one as draft meanwhile

flomnes (approved):

PR #898, opened 2024-08-12
Test assets are running in fresh environnement (CentOS7) 

JasonMarechal25:


.github/workflows/build_centos7.yml
```diff
@@ -183,16 +183,6 @@ jobs:
           path: ${{ env.VCPKG_CACHE_DIR }}
           key: vcpkg-cache-centOS-${{ hashFiles('src/vcpkg.json', '.git/modules/vcpkg/HEAD') }}
 
-      - name: Running unit tests
-        timeout-minutes: 120
-        shell: bash
-        run: |
-          export PATH=${GITHUB_WORKSPACE}/_build/vcpkg_installed/x64-linux/tools/openmpi/bin/:$PATH
-          export LD_LIBRARY_PATH=LD_LIBRARY_PATH:${{ env.XPRS_LIB_Path_CONTAINER }}
-          export XPRESS=${{ env.XPRESS_CONTAINER }}
```
flomnes: How come the new scripts don't source XPRESS ?

.github/workflows/build_centos7.yml
```diff
@@ -262,6 +252,119 @@ jobs:
...
+        name: run tests with 2 processes
+        with:
+          image: centos:run
+          shell: bash
+          options: -v ${{ github.workspace }}:/work
+          run: |
+            export PATH=$PATH:/usr/lib64/openmpi/bin
+            export LD_LIBRARY_PATH=$LD_LIBRARY_PATH:/usr/lib64/openmpi/lib
+            cd /work
+            ./antares-xpansion-launcher -i examples/SmallTestFiveCandidates -n 2 --allow-run-as-root
```
flomnes, line 376: You could merge these two into the same step

```suggestion
      - uses: addnab/docker-run-action@v3
        name: Run tests
        with:
          image: centos:run
          shell: bash
          options: -v ${{ github.workspace }}:/work
          run: |
            export PATH=$PATH:/usr/lib64/openmpi/bin
            export LD_LIBRARY_PATH=$LD_LIBRARY_PATH:/usr/lib64/openmpi/lib
            cd /work
            ./antares-xpansion-launcher -i examples/SmallTestFiveCandidates --allow-run-as-root
            ./antares-xpansion-launcher -i examples/SmallTestFiveCandidates -n 2 --allow-run-as-root
```

.github/workflows/build_centos7.yml
```diff
@@ -262,6 +252,119 @@ jobs:
...
```
flomnes, line 329: You could merge these two steps into one

.github/workflows/build_centos7.yml
```diff
@@ -262,6 +252,119 @@ jobs:
...
+
+      - name: Download image
+        uses: actions/download-artifact@v4
+        with:
+          name: centos
+          path: /tmp
+
+      - name: Load image
+        run: |
+          docker load --input /tmp/centos.tar
```
flomnes, line 348: You may add a some compression (.tar.gz) to save space (on runners & in cache), gzip doesn't cost much CPU time.

.github/workflows/build_centos7.yml
JasonMarechal25: Mistake on my side. I've forgotten to remove this deletion. It was to disable test run during development

.github/workflows/build_centos7.yml
JasonMarechal25: If you're talking about the two steps  _Run tests_ and run _tests with 2 processes_ it is to better distinguish what failed in case of an error. One, the other, or both

If you're talking about the two jobs _test_assets_ and _test_single_file_asset_ same things but also jobs are run in parallel

.github/workflows/build_centos7.yml
JasonMarechal25: See my comment for justification: https://github.com/AntaresSimulatorTeam/antares-xpansion/pull/898/files#r1743331998

.github/workflows/build_centos7.yml
JasonMarechal25: _upload artifact_ action already compress assets by default.
Unfortunately the content of centos.tar is blob and doesn't compress well

flomnes (approved):

PR #909, opened 2024-09-03
fix dangling ref 

a-zakir:


JasonMarechal25 (approved):

a-zakir:
used [here](https://github.com/AntaresSimulatorTeam/antares-xpansion/blob/22f6c1d7887d90cffbfaa521f2b9861c6117f630/src/cpp/benders/outer_loop/OuterLoop.cpp#L25) for OuterLoop::outer_loop_biLevel_ (which is OuterLoopBiLevel)

PR #929, opened 2024-09-12
fix warnings

a-zakir:


flomnes (commented):
Not sure you need `= ""` for `std::string`s and `= {}` for `std::vector`s

a-zakir:
> Not sure you need `= ""` for `std::string`s and `= {}` for `std::vector`s

they are maybe unnecessary, it doesn't cost anything to do so though

flomnes (approved):

PR #941, opened 2024-10-07
ANT-2223

a-zakir:
##
- [x] search unsp variables for all areas
- [x] compute LOLD 
- [ ] ~need more details on candidate capacity / area~


src/cpp/benders/benders_core/include/BendersBase.h
```diff
@@ -227,6 +229,12 @@ class BendersBase {
                                PlainData::SubProblemData &subproblem_data,
                                const std::string &name,
                                const std::shared_ptr<SubproblemWorker> &worker);
+  // TODO to be rethink
+  std::shared_ptr<Outerloop::CriterionComputation> criterions_computation_;
```
tbittar: It seems weird to have a reference to `Outerloop` namespace within the `BendersBase` class....

src/cpp/benders/benders_mpi/include/BendersMPI.h
```diff
@@ -87,8 +87,14 @@ class BendersMpi : public BendersBase {
...
   virtual void GatherCuts(const SubProblemDataMap &subproblem_data_map,
                           const Timer &walltime);
+  void BroadCastVariablesIndices();
+  virtual void ComputeSubproblemsContributionToOuterLoopCriterion(
+      const SubProblemDataMap &subproblem_data_map);
+  void SolveSubproblem(
+      SubProblemDataMap &subproblem_data_map,
+      PlainData::SubProblemData &subproblem_data, const std::string &name,
+      const std::shared_ptr<SubproblemWorker> &worker) override;
+  void UpdateOuterLoopMaxCriterionArea();
```
tbittar: I have the feeling the the outer loop is now mixed with Benders, and the fact that the outer loop is a loop around Benders becomes unclear when reading the code

tbittar: The thing is that the "criterion" can refer to something that is not specific to the outer loop, so maybe the notion of criterion is not attached to the outer loop and refers to something more general

src/cpp/benders/factories/BendersFactory.cpp
```diff
@@ -34,154 +28,237 @@ BENDERSMETHOD DeduceBendersMethod(size_t coupling_map_size, size_t batch_size,
...
+}
+
+Outerloop::OuterLoopInputData BendersMainFactory::PatternsFromSupbProblem(
+    const std::string& first_subproblem_name) const {
+  SolverAbstract::Ptr solver = BuildSolver(first_subproblem_name);
+  const auto all_variables_name = solver->get_col_names();
+  std::set<std::string> unique_areas = UniqueAreas(all_variables_name);
+  Outerloop::OuterLoopInputData ret;
+  // TODO check this value
+  ret.SetCriterionCountThreshold(1);
```
tbittar, line 211: It's OK !

src/cpp/benders/factories/BendersFactory.cpp
```diff
@@ -34,154 +28,237 @@ BENDERSMETHOD DeduceBendersMethod(size_t coupling_map_size, size_t batch_size,
...
+  }
+  return unique_areas;
+}
+SolverAbstract::Ptr BendersMainFactory::BuildSolver(
+    const std::string& first_subproblem_name) const {
+  SolverFactory factory(logger_);
+  auto solver_log_manager = SolverLogManager(LogReportsName());
+  auto solver = factory.create_solver(
+      options_.SOLVER_NAME, SOLVER_TYPE::CONTINUOUS, solver_log_manager);
+  solver->set_threads(1);
```
tbittar: Put `set_threads(1)` after `read_prob_mps`, this will avoid us some bugs later on when reading from binary saved files...

tbittar (changes requested):

src/cpp/benders/benders_core/include/BendersBase.h
a-zakir: I agree

a-zakir: do you want me to rename/move this object?

src/cpp/benders/benders_mpi/include/BendersMPI.h
a-zakir: what do you think on renaming:
UpdateOuterLoopMaxCriterionArea -> UpdateMaxCriterionArea ?

src/cpp/benders/benders_core/include/BendersBase.h
tbittar: Yes I think that the criterion computation is a concept that is not necessarily linked to the outer loop as we can resort to it within Benders just to have outputs. However the use in Benders and in outer loop may have different objectives as in Benders it is just for outputs, whereas in outer loop, it is used to within the algo (to update the lambda), so I do not know if keeping a single concept is enough or if we have to split between something used to compute any outputs requested by the user and something to be used in the outer loop algorithm ? What do you think ?

tbittar (approved):

PR #944, opened 2024-10-08
Fix generation of weights in API mode [ANT-2196]

pet-mit:
When generating weights.txt, instead of using list of MPS files generated by simulator in the archive, use the list of MPS files generated by XPANSION in output_dir/lp.
This makes it compatible with API mode.

Also added a cucumber test with a new step to launch xpansion (and fixed the "simulation succeeds" step)

### DIFF
--- a/src/cpp/lpnamer/input_reader/WeightsFileWriter.cpp
+++ b/src/cpp/lpnamer/input_reader/WeightsFileWriter.cpp
@@ -2,6 +2,7 @@
 
 #include <fstream>
 #include <numeric>
+#include <ranges>
 
 #include "StringManip.h"
 
@@ -27,16 +28,16 @@ void YearlyWeightsWriter::CreateWeightFile() {
 
 void YearlyWeightsWriter::FillMpsWeightsMap() {
   mps_weights_.clear();
-  std::vector<std::filesystem::path> mps_files;
-  for (auto& p : std::filesystem::directory_iterator(xpansion_lp_dir_)) {
-    if (p.path().extension() == ".mps") mps_files.emplace_back(p.path());
-  }
-  for (auto& mps_file : mps_files) {
-    auto year = GetYearFromMpsName(mps_file.filename().string());
+  auto it = std::filesystem::directory_iterator(xpansion_lp_dir_);
+  auto mps_list = it | std::views::filter([](const auto& p) {
+                    return p.path().extension() == ".mps";
+                  });
+  for (auto& mps_file : mps_list ) {
+    auto year = GetYearFromMpsName(mps_file.path().filename().string());
     auto year_index =
         std::find(active_years_.begin(), active_years_.end(), year) -
         active_years_.begin();
-    mps_weights_[mps_file.filename()] = weights_vector_[year_index];
+    mps_weights_[mps_file.path().filename()] = weights_vector_[year_index];
   }
 }
 
--- a/src/cpp/lpnamer/input_reader/WeightsFileWriter.cpp
+++ b/src/cpp/lpnamer/input_reader/WeightsFileWriter.cpp
@@ -3,17 +3,14 @@
 #include <fstream>
 #include <numeric>
 
-#include "ArchiveReader.h"
 #include "StringManip.h"
 
 YearlyWeightsWriter::YearlyWeightsWriter(
     const std::filesystem::path& xpansion_output_dir,
-    const std::filesystem::path& antares_archive_path,
     const std::vector<double>& weights_vector,
     const std::filesystem::path& output_file,
     const std::vector<int>& active_years)
     : xpansion_output_dir_(xpansion_output_dir),
-      antares_archive_path_(antares_archive_path),
       weights_vector_(weights_vector),
       output_file_(output_file),
       active_years_(active_years) {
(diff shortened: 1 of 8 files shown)


src/cpp/lpnamer/input_reader/WeightsFileWriter.cpp
```diff
@@ -30,19 +27,17 @@ void YearlyWeightsWriter::CreateWeightFile() {
 
 void YearlyWeightsWriter::FillMpsWeightsMap() {
   mps_weights_.clear();
-  auto zip_reader = ArchiveReader(antares_archive_path_);
-  zip_reader.Open();
-  zip_reader.LoadEntriesPath();
-  const auto& mps_files = zip_reader.GetEntriesPathWithExtension(".mps");
+  std::vector<std::filesystem::path> mps_files;
```
JasonMarechal25: We can do with a filtered view instead of an intermediate vector:
```
 auto mps_list = std::filesystem::directory_iterator(xpansion_lp_dir_) | std::views::filter([](const auto& p) {
                    return p.path().extension() == ".mps";
                  });
  for (auto& mps_file : mps_list ) {
    auto year = GetYearFromMpsName(mps_file.path().filename().string());
    auto year_index =
        std::find(active_years_.begin(), active_years_.end(), year) -
        active_years_.begin();
    mps_weights_[mps_file.path().filename()] = weights_vector_[year_index];
  }
```

JasonMarechal25 (changes requested):

src/cpp/lpnamer/input_reader/WeightsFileWriter.cpp
pet-mit: it seems we cannot apply filters this way on directory_iterator, have you tried it ?

src/cpp/lpnamer/input_reader/WeightsFileWriter.cpp
```diff
@@ -30,19 +27,17 @@ void YearlyWeightsWriter::CreateWeightFile() {
...
+  for (auto& p : std::filesystem::directory_iterator(xpansion_lp_dir_)) {
+    if (p.path().extension() == ".mps") mps_files.emplace_back(p.path());
```
a-zakir: Please use braces in the “if” condition to keep the same format/style as the code.

tests/end_to_end/cucumber/steps/steps.py
```diff
@@ -82,3 +95,10 @@ def assert_dict_allclose(actual, expected, rtol=1e-06, atol=0):
 def check_solution(context):
     expected_solution = {row['variable']: float(row['value']) for row in context.table}
     assert_dict_allclose(context.outputs["solution"]["values"], expected_solution)
+
+
+def parse_output_folder_from_logs(logs: bytes) -> str:
+    for line in logs.splitlines():
+        if b'Output folder : ' in line:
+            return line.split(b'Output folder : ')[1].decode('ascii')
```
a-zakir: i guess you meant :"Optimization results available "

tests/end_to_end/cucumber/steps/steps.py
```diff
@@ -51,20 +56,28 @@ def run_outer_loop(context, n):
...
 
 
-@then("the simulation takes less than {seconds} seconds")
+@when('I run antares-xpansion in memory with the {method} method and {n:d} proc(s)')
+def run_antares_xpansion(context, method, n):
+    command = build_launch_command(context.study_path, method, n, True)
+    process = subprocess.Popen(command, stdout=subprocess.PIPE, stderr=subprocess.DEVNULL, shell=True)
+    out, err = process.communicate()
+    context.return_code = process.returncode
+    context.outputs = read_outputs(Path(parse_output_folder_from_logs(out)) / "expansion" / "out.json")
```
a-zakir, line 65: why  bother with parsing logs? --> read_outputs(Path("..") / "expansion" / "out.json")
+ the output file is not logged in all steps

src/cpp/lpnamer/input_reader/WeightsFileWriter.cpp
JasonMarechal25: Yes. But indeed it doesn't work in this form. However it works with an an intermediate value:
```
 auto it = std::filesystem::directory_iterator(xpansion_lp_dir_);
 auto mps_list = it | std::views::filter([](const auto& p) {
                    return p.path().extension() == ".mps";
                  });
  for (auto& mps_file : mps_list ) {
    auto year = GetYearFromMpsName(mps_file.path().filename().string());
    auto year_index =
        std::find(active_years_.begin(), active_years_.end(), year) -
        active_years_.begin();
    mps_weights_[mps_file.path().filename()] = weights_vector_[year_index];
  }
```

tests/end_to_end/cucumber/features/weights.feature
```diff
@@ -0,0 +1,9 @@
+Feature: add weights on MC years
+
+  @slow @short @full-launch
+  Scenario: handling different weights on MC years in API mode
+    # For now, only check that the simulation succeeds
+    # TODO : add more non-regression tests when we have more steps
+    Given the study path is "examples/SmallTestFiveCandidatesWithWeights"
```
JasonMarechal25:
```suggestion
    Given the study path is "examples/SmallTestFiveCandidatesWithWeights" for a study with weights
```

tests/end_to_end/cucumber/steps/steps.py
pet-mit: you're right, I copied what I had done for simulator, but it's not right. yours is better

a-zakir (approved):

JasonMarechal25 (approved):

tests/end_to_end/cucumber/features/weights.feature
pet-mit: since we don't need to change the step definition (code-behind), changing the step description has no real advantage IMO

JasonMarechal25: It was for clarity. The goal of the test is to test with weight, the study is not interchangeable with any other.

PR #947, opened 2024-10-14
Target to generate graph

JasonMarechal25:
Add a "graphviz" target to generate dot, png and svg files representing CMake dependency graph.


CMakeLists.txt
```diff
@@ -407,3 +407,10 @@ endif ()
 #version file to define Xpansion and Antares-simulator versions
 configure_file(${CMAKE_CURRENT_SOURCE_DIR}/src/python/antares_xpansion/__version__.py.in ${CMAKE_CURRENT_SOURCE_DIR}/src/python/antares_xpansion/__version__.py)
 include(CPack)
+
```
a-zakir, line 410: this should not be generated automatically!

a-zakir: only on demand and surely not in the cmake; I think this belongs to doc

JasonMarechal25: It is on demand, to have it at every configure you would need to add "ALL" to add_custom_target.
This commands takes to much time however to be run systemically at configure or build. 
Here we only define the target for users (developers) to run manually

a-zakir (approved):

PR #950, opened 2024-10-14
Avoid computing max criterion when there is none

a-zakir:


tests/end_to_end/cucumber/steps/steps.py
```diff
@@ -97,6 +107,31 @@ def check_solution(context):
...
+        next(file)
+
+        # Check each line in the file
+        for line in file:
+            # Split the line by whitespace and get the value in the target column
+            columns = line.split()
+
+            # Check if the column value is not zero
+            try:
+                if float(columns[column_index]) != 0.0:
```
tbittar: Float equality comparison is dangerous, use `math.isclose`

tbittar (changes requested):

tests/end_to_end/cucumber/steps/steps.py
```diff
@@ -107,23 +108,30 @@ def check_solution(context):
...
             try:
-                if float(columns[column_index]) != 0.0:
-                    return False
+                value = float(columns[column_index])
             except (ValueError, IndexError):
                 print(f"Error parsing line: {line.strip()}")
                 return False
+
+            # Use math.isclose to compare to zero with tolerance
+            if not math.isclose(value, 0.0, rel_tol=rel_tol, abs_tol=abs_tol):
```
tbittar: As you compare a value to 0, `rel_tol` is useless, you should rely on `abs_tol`, you can take `10^-9` for instance

tbittar (changes requested):

tests/end_to_end/cucumber/steps/steps.py
a-zakir: done

tbittar (approved):

PR #959, opened 2024-10-29
[ANT-2299] Enable save/restore with xpress and enable by default (ProblemGeneration)

JasonMarechal25:
Expend solver API with save_prob and restore_prob. Mirror Xpress API to save and restore problems

Coin solvers lack save/restore options. Using those method default to writing mps files

- New Benders option to pilot if we need to use save/restore or read/write[mps]
- Xpansion use Save/restore by default.
- ProblemGeneration use save/restore by default. When using xpress it'll work as expected. When using Coin it'll fallback to writing MPS

Note: 

- The scope of this PR is limited to the frontier between problemGeneration and Benders
- For lpnamer e2e test with xpress, we only test the existence of a .svf file. We rely on xpansion e2e test for correctness. Having to develop a mean to check correctness of svf is time consuming and svf support is just an intermediate step and will be retired in a short horizon of time.
- Benders should still support mps files as input for all solver (r&d use case)


src/cpp/multisolver_interface/environment.cc
```diff
@@ -45,6 +45,13 @@ std::function<int(XPRSprob prob, const char* filename, const char* flags)>
     XPRSwritebasis = nullptr;
 std::function<int(XPRSprob prob, const char* filename, const char* flags)>
     XPRSreadprob = nullptr;
+std::function<int(XPRSprob prob, int nr, int nc, const int mrow[],
```
tbittar, line 48: Not related to this PR, but needed later for presolve

src/cpp/lpnamer/model/Problem.cpp
```diff
@@ -10,3 +10,9 @@ void Problem::read_prob_mps(const std::filesystem::path& filename) {
   std::tie(mc_year, week) = McYearAndWeek(filename);
   solver_abstract_->read_prob_mps(filename);
 }
+void Problem::save_prob(const std::filesystem::path& filename) {
```
tbittar, line 13: Do we need to do `std::tie(mc_year, week) = McYearAndWeek(filename);` similarly as in the previous function ?

tbittar (commented):
Note that when writing the master problem at the end of each iteration, we still use `write_prob_mps` (which is further read in sensitivity analysis with `read_prob_mps`), should we keep it like this or use save/restore as well here ?

src/cpp/lpnamer/model/Problem.cpp
JasonMarechal25: Maybe, will need to check

tests/cpp/full_run/FullRunProblemPassingTest.cpp
```diff
@@ -0,0 +1,5 @@
+#include "gtest/gtest.h"
```
JasonMarechal25: Need to remove

src/cpp/lpnamer/model/Problem.cpp
JasonMarechal25: For save we don't need to.
However for restore we do, it's the same as read.
good catch

src/cpp/lpnamer/main/ProblemGeneration.cpp
```diff
@@ -320,8 +323,8 @@ void ProblemGeneration::RunProblemGeneration(
   LinkProblemsGenerator linkProblemsGenerator(
       lpDir_, links, solver_config_, logger, solver_log_manager, rename_problems);
   std::shared_ptr<ArchiveReader> reader =
-      antares_archive_path.empty() ? std::make_shared<ArchiveReader>()
-                                   : InstantiateZipReader(antares_archive_path);
+      mode_ == SimulationInputMode::ARCHIVE ? InstantiateZipReader(antares_archive_path)
```
tbittar, line 326: Que fait cette modif ?

src/cpp/lpnamer/problem_modifier/include/antares-xpansion/lpnamer/problem_modifier/MPSFileProblemProviderAdapter.h
```diff
@@ -5,6 +5,7 @@
 #pragma once
 
 #include "antares-xpansion/lpnamer/problem_modifier/IProblemProviderPort.h"
+#include "antares-xpansion/multisolver_interface/SolverConfig.h"
```
tbittar: Unused ? Maybe for later on ?

src/cpp/lpnamer/problem_modifier/MasterGeneration.cpp
```diff
@@ -36,32 +45,40 @@ void MasterGeneration::write_master_mps(
...
+  Problem master_problem(master_l);
+  master_problem._name = "master";
+  master_problem.save_prob(rootPath / "lp" / "master");
+}
 
-  std::string const &lp_name = "master";
-  master_l->write_prob_mps(rootPath / "lp" / (lp_name + ".mps"));
+std::filesystem::path FileNameForStructureFile(const std::string& problemName,
+                                             std::string solverName) {
+  const std::string save_ext = ".svf";
```
tbittar: Seems unused, as moved to `FileName`

src/cpp/lpnamer/problem_modifier/MasterGeneration.cpp
```diff
@@ -36,32 +45,40 @@ void MasterGeneration::write_master_mps(
...
+  const std::string default_ext = ".mps";
```
tbittar: Seems unused

src/cpp/multisolver_interface/SolverConfig.cpp
```diff
@@ -2,6 +2,7 @@
 
 #include <algorithm>
 #include <utility>
+#include "antares-xpansion/xpansion_interfaces/LogUtils.h"
```
tbittar, line 5: Useless

src/cpp/multisolver_interface/SolverFactory.cpp
```diff
@@ -9,22 +11,48 @@
...
+
+std::vector<std::string> SolverLoader::GetAvailableSolvers(
+    std::shared_ptr<ILoggerXpansion> logger) {
+  std::call_once(solver_flag, GetAvailableSolversInternal, logger);
   return available_solvers;
 }
+std::vector<std::string> SolverLoader::GetSupportedSolvers() {
+  static std::vector<std::string> supported_solvers;
+  if (supported_solvers.empty()) {
+    supported_solvers.emplace_back(XPRESS_STR);
```
tbittar, line 54: I don't understand what happens here, it looks like Xpress will always be in the supported_solver list

src/python/antares_xpansion/xpansionConfig.py
```diff
@@ -167,6 +167,7 @@ def _set_default_settings(self):
             "log_level": "0",
             "separation_parameter": "0.5",
             "batch_size": "0",
+            "problems_format": "SAVED",
```
tbittar, line 170: The default value is ok even for open source solvers, as they will replace .svf to .mps and write files anyway, am I correct ?

JasonMarechal25: yes

src/cpp/multisolver_interface/SolverFactory.cpp
JasonMarechal25: Ok, I'm adding a comment. Support is "I know such solver exists" it differs from "available" because it doesn't check for licence for example.
Currently the only usage is for testing (buty tricky to move in tests)

src/cpp/lpnamer/main/ProblemGeneration.cpp
JasonMarechal25: Juste de la clarté. Pour savoir si on es en mode archive c'est plus propre de tester le mode_ que si le chamin vers l'archive est renseigné

src/cpp/multisolver_interface/SolverConfig.cpp
JasonMarechal25: It is not. It defines the type of exceptions used in the file

tbittar (approved):

PR #964, opened 2024-11-08
[ANT-1226] Cohérence des solver E2E

JasonMarechal25:
Ensure the same solver is used across all steps of Xpansion:

- Use the solver described in user/expansion/settings for Antares Simulator
- Use the solver described in user/expansion/settings for ProblemGeneration
- No change to Benders

Limitation: Solver is not passed to Antares Simulator in case of "in-memory" mode.


src/cpp/lpnamer/input_reader/SettingsReader.cpp
```diff
@@ -0,0 +1,54 @@
...
+    }
+    this->settings_[StringManip::trim(values[0])] = StringManip::trim(values[1]);
+  }
+}
+void SettingsReader::check_file_exist(
+    const std::filesystem::path& file_path) const {
+  if (file_path.empty() || !std::filesystem::exists(file_path)) {
+    std::ostringstream msg;
+    msg << LOGLOCATION
+        << "General data file is not found : " << file_path.string()
```
tbittar: Settings file is not found (general data refers to Antares Simulator)

src/cpp/lpnamer/main/ProblemGeneration.cpp
```diff
@@ -111,6 +113,9 @@ std::filesystem::path ProblemGeneration::updateProblems() {
   auto weights_file = options_.WeightsFile();
   auto unnamed_problems = options_.UnnamedProblems();
 
+  SettingsReader settingsReader(xpansion_output_dir, logger.get());
```
tbittar: I don't understand how this works : you give xpansion output directory but then it does `check_file_exist` but you have only given the directory, I do not see where the path of the settings file is retrieved (probably I missed something in the code...), and why the end-to-end tests do not raise an `IniFileNotFound` exception

JasonMarechal25: Indeed a mistake from me.

a-zakir:
putting the solver name in a file is good, i would also include all pb gen options, what do you think?

tbittar (approved):

PR #968, opened 2024-11-27
Enable OpenMP compile flags for std::execution_policy::par support

flomnes:


tbittar:
Does not fix the problem with parallelisation with execution policies, we need to restore dependency with TBB in the short term

PR #970, opened 2024-11-29
add TBB to vcpkg

a-zakir:


vcpkg.json
```diff
@@ -33,6 +33,13 @@
       "features": [
         "zlib"
       ]
+    },
+    "tbb"
+  ],
+  "overrides": [
+    {
+      "name": "tbb",
+      "version-string": "2020_U3"
```
flomnes, line 42: Why is this needed ?

flomnes (commented):
Why not one global `find_package(TBB <options>)` at the root CMakeLists.txt ?

vcpkg.json
a-zakir: gcc 10

pet-mit (approved):

JasonMarechal25:
@a-zakir Can you expand why TBB need to be reintroduced? 
TBB was introduced as a backend for parallel execution policy: https://en.cppreference.com/w/cpp/algorithm/execution_policy_tag_t
As per gcc10, execution policy should be supported by libstdc++ and shouldn't require TBB anymore.

a-zakir:
> @a-zakir Can you expand why TBB need to be reintroduced? TBB was introduced as a backend for parallel execution policy: https://en.cppreference.com/w/cpp/algorithm/execution_policy_tag_t As per gcc10, execution policy should be supported by libstdc++ and shouldn't require TBB anymore.

i don't know, i was told to do it, to restore lost perf

PR #971, opened 2024-12-06
Fix/centos build

a-zakir:


docker/centos7-build
```diff
@@ -0,0 +1,128 @@
...
+
+RUN mkdir -p /workspace/vcpkg_cache
+ARG VCPKG_CACHE_DIR
+COPY $VCPKG_CACHE_DIR /workspace/
+
+
+ARG BRANCH=develop
+RUN cd /workspace && \
+    source /opt/rh/rh-git227/enable && \
+    git clone https://github.com/AntaresSimulatorTeam/antares-xpansion.git --branch $BRANCH && \
```
flomnes:
```suggestion
    git clone https://github.com/AntaresSimulatorTeam/antares-xpansion.git --depth 1 --branch $BRANCH && \
```

flomnes: Use shallow clone to reduce bandwidth & disk usage 🍃

docker/centos7-build
```diff
@@ -0,0 +1,128 @@
...
+    wget $URL_ANTARES && \
+    tar -xvf antares-${ANTARES_VERSION}-CentOS-7.9.2009.tar.gz -C deps --strip-components=1 &&\
+    rm -rf antares-${ANTARES_VERSION}-CentOS-7.9.2009.tar.gz
+
+WORKDIR /workspace/antares-xpansion
+
+RUN pip3 install --upgrade pip && \
+    pip3 install wheel && \
+    pip3 install -r requirements-tests.txt && \
+    pip3 install -r requirements-ui.txt
```
flomnes, line 50: pip [accepts multiple arguments](https://stackoverflow.com/a/36744306)

```suggestion
    pip3 install -r requirements-tests.txt -r requirements-ui.txt
```

flomnes (commented):
LGTM, a few remarks

flomnes:
Also there seems to be an issue with PyQT5 being incompatible with python3.9 on Windows & Ubuntu, so you have to take care of this too

a-zakir:
> Also there seems to be an issue with PyQT5 being incompatible with python3.9 on Windows & Ubuntu, so you have to take care of this too

unrelated to this pr, in fact the centos build succeed

docker/centos7-build
a-zakir: good to know

flomnes (approved):

PR #977, opened 2024-12-20
Use updated antares-solver API (output path changes) [ANT-2561]

flomnes:
### NOTE
This PR depends on https://github.com/AntaresSimulatorTeam/Antares_Simulator/pull/2548 because the signature of some functions were changed (output directory becomes a return value).

### DIFF
--- a/src/cpp/lpnamer/main/ProblemGeneration.cpp
+++ b/src/cpp/lpnamer/main/ProblemGeneration.cpp
@@ -61,10 +61,19 @@ ProblemGeneration::ProblemGeneration(ProblemGenerationOptions& options)
   }
 }
 
-std::filesystem::path ProblemGeneration::performAntaresSimulation() {
+static std::string lowerCase(std::string data) {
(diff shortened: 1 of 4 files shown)


src/cpp/lpnamer/main/ProblemGeneration.cpp
```diff
@@ -61,10 +61,19 @@ ProblemGeneration::ProblemGeneration(ProblemGenerationOptions& options)
   }
 }
 
-std::filesystem::path ProblemGeneration::performAntaresSimulation() {
+static std::string lowerCase(std::string data) {
+  std::transform(data.begin(), data.end(), data.begin(),
+                 [](unsigned char c) { return std::tolower(c); });
+  return data;
+}
```
flomnes, line 68: Simulator only accepts e.g `xpress`, Xpansion only accepts e.g `xpress`. We just lowercase the input before giving it to Simulator.

tbittar (approved):

PR #986, opened 2025-01-15
[ANT-3671] Fix numerical issues in Benders

tbittar:
Several numerical issues can occur in Benders
- 
- [x] Fixing candidate values in the subproblems to the value obtained from the master can lead to subproblem infeasibility if the master solution is numerically slightly below the lb, above the ub, or is a not rounded integer value. Also master solution that are close to zero given a tolerance should be set to zero in the subproblems to avoid having large range of coefficients in the subproblems leading to numerical issues.
- [x] Subproblem gradients that are close to zero (given a tolerance) should appear as zero in the cut added to the master to avoid numerical issues in the master due to large range of coefficients values


hugo-digiugno:
Hey! Any idea on when this fix will be available as part of a new release? Thanks in advance

tbittar:
Hello ! I do not know yet as this work is still at R&D stage

tbittar:
@JasonMarechal25 several points of attention during the review :

- I have added problem providers for the workers, but maybe the API can be improve, especially I have a method `provide_file_path` only used to build file path to write the problem in case of error. Maybe it is not the responsibility of this provider ?
- Some public methods have been added to WorkerMaster https://github.com/AntaresSimulatorTeam/antares-xpansion/blob/313bc936cb5d8266da8ed9bdb82021cb473b4deb/src/cpp/benders/benders_core/include/antares-xpansion/benders/benders_core/WorkerMaster.h#L80-L92 only for the need of tests in `WorkerMasterTest`. The design maybe improved
- There is a memory error when running `WorkerMasterTest` that needs to be fixed

JasonMarechal25 (approved):

PR #990, opened 2025-01-28
Weights not correctly read for subproblems

tbittar:
In problem generation, a file `weights.txt` is generated to map problem names to their corresponding weights. Initially, this mapping was constructed by fetching all mps files in the directory after the antares step and extracting the year number from the file name.

In memory mode, Antares does not write mps file anymore. The weights file generation is now based on the `problems_and_data` structure in memory that gathers the problems and their year number. Moreover, depending on the solver name, the filename written in the `weights.txt` is now correct (.svf or .mps) depending on the context.

### DIFF
--- a/src/cpp/benders/benders_core/Worker.cpp
+++ b/src/cpp/benders/benders_core/Worker.cpp
@@ -49,7 +49,7 @@ void Worker::init(const std::string &solver_name, int log_level,
   }
 
   read_prob(_solver.get(), _path_to_mps);
-  // Always set solver parameters after reading problems, as restore also comes with parameters of the solver and we do not want them to override our preferences
+  // Always set solver parameters after reading problems, as restore (used by Xpress writing .svf files) also comes with parameters of the solver and we do not want them to override our preferences
   _solver->set_threads(1);
   _solver->set_output_log_level(log_level);
 
--- a/src/cpp/benders/benders_core/Worker.cpp
+++ b/src/cpp/benders/benders_core/Worker.cpp
@@ -48,9 +48,10 @@ void Worker::init(const std::string &solver_name, int log_level,
                                     solver_log_manager);
   }
 
+  read_prob(_solver.get(), _path_to_mps);
+  // Always set solver parameters after reading problems, as restore also comes with parameters of the solver and we do not want them to override our preferences
   _solver->set_threads(1);
   _solver->set_output_log_level(log_level);
-  read_prob(_solver.get(), _path_to_mps);
 
   for (auto const &kvp : _name_to_id) {
     _id_to_name[kvp.second] = kvp.first;
--- a/src/cpp/lpnamer/input_reader/WeightsFileWriter.cpp
+++ b/src/cpp/lpnamer/input_reader/WeightsFileWriter.cpp
@@ -1,5 +1,7 @@
 #include "antares-xpansion/lpnamer/input_reader/WeightsFileWriter.h"
 
+#include <antares-xpansion/multisolver_interface/SolverConfig.h>
+
 #include <fstream>
 #include <numeric>
 #include <ranges>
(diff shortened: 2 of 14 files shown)


src/cpp/lpnamer/input_reader/WeightsFileWriter.cpp
```diff
@@ -11,34 +13,35 @@ YearlyWeightsWriter::YearlyWeightsWriter(
...
-    auto year = GetYearFromMpsName(mps_file.path().filename().string());
-    auto year_index =
-        std::find(active_years_.begin(), active_years_.end(), year) -
-        active_years_.begin();
-    mps_weights_[mps_file.path().filename()] = weights_vector_[year_index];
+
+  for (const auto& [problem, data] : problems_and_data) {
+    auto year_index = std::find(active_years_.begin(), active_years_.end(),
+                                problem->mc_year) -
+                      active_years_.begin();
```
flomnes: ⚠️ This line is really bad

`std::find` returns an iterator to the first element in the range [first, last) that satisfies specific criteria (or last if there is no such iterator). 

`std::find` could fail to find a matching element, in which case it would return `active_years_.end()`. This case is not handled and lead to an invalid `year_index`.

src/cpp/lpnamer/input_reader/WeightsFileWriter.cpp
```diff
@@ -11,34 +13,35 @@ YearlyWeightsWriter::YearlyWeightsWriter(
...
+    mps_weights_[data._problem_mps] = weights_vector_[year_index];
```
flomnes: The name `mps_weights_` is misleading, especially in case we don't read MPS (in-memory mode)

flomnes: The name is misleading especially in case we don't read MPS (in-memory mode)

src/cpp/lpnamer/input_reader/WeightsFileWriter.cpp
```diff
@@ -54,7 +57,10 @@ void YearlyWeightsWriter::DumpMpsWeightsToFile() const {
   mps_weights_file.open(file);
 
   for (const auto& [mps_name, weight] : mps_weights_) {
-    mps_weights_file << mps_name.string() << " " << weight << std::endl;
+    mps_weights_file << SolverConfig::FileName(
+                            mps_name, SolverConfig(std::move(solver_name_)))
```
flomnes: Careful with `std::move`, it invalidates it's argument
```suggestion
                            mps_name, SolverConfig(solver_name_))
```

src/cpp/benders/benders_core/Worker.cpp
```diff
@@ -48,9 +48,10 @@ void Worker::init(const std::string &solver_name, int log_level,
                                     solver_log_manager);
   }
 
+  read_prob(_solver.get(), _path_to_mps);
+  // Always set solver parameters after reading problems, as restore also comes with parameters of the solver and we do not want them to override our preferences
```
flomnes: Mention xpress & .svf files

src/cpp/lpnamer/input_reader/WeightsFileWriter.cpp
flomnes: Method `SolverConfig::FileName` can surely be improved

```cpp
// Current : use another instance
static std::filesystem::path SolverConfig::FileName(const SolverConfig& c);
// Proposed : use *this
std::filesystem::path SolverConfig::FileName() const;
```

src/cpp/lpnamer/input_reader/WeightsFileWriter.cpp
flomnes: Also, you convert a position in a vector to a year. This isn't very robust since the order could change in the future, there could be missing elements, etc.

tbittar: No `year_index` is not a year, it is a position in a vector... this line already exists in the code, it is not new by the way (even though I agree it can be more robust)

tbittar: Existing `problem->mc_year` are exactly the active years of the study, and we need to retrieve the index of the current mc year in the list of active years to read the correct line of the weights file. I do not know how we can do this much better (except by changing the data structure of `active_years` to store a tuple (year_index, year) which may be overkill)

tbittar: Log message added in case `mc_year` not found in `active_years`

src/cpp/lpnamer/input_reader/WeightsFileWriter.cpp
tbittar: Names updated

src/cpp/lpnamer/input_reader/WeightsFileWriter.cpp
tbittar: Suggestion applied

src/cpp/benders/benders_core/Worker.cpp
tbittar: Done

flomnes (approved):

PR #991, opened 2025-01-29
Remove Antares mps file after read in problem generation

tbittar:
Antares files are read in problem generation and are then useless, they should be removed to improve disk usage


flomnes (commented):
[`std::remove`](https://en.cppreference.com/w/cpp/io/c/remove) is a C function from defined in header stdio.h, it takes a `const char*` and returns an `int`
On the other hand [`std::filesystem::remove`](https://en.cppreference.com/w/cpp/filesystem/remove) takes a `const std::filesystem::path&` and returns a `bool`

Even though both these functions appear to do the same thing, I think you should use the 2nd one as it's more natural and won't implicitly convert `std::filesystem::path` -> `const char*`.

In any case, I think you should log a warning in case the file couldn't be removed (check the return code)

flomnes (approved):

PR #997, opened 2025-02-06
Update and apply Clang format

JasonMarechal25:
Update clang-format with format used in Antares Simulator for homogeneity
Add job to check format


format-code.sh
```diff
@@ -0,0 +1,24 @@
+#!/bin/bash
+
+if [ $# -eq 0 ]
+then
+    # No arguments: format all
+    SOURCE_DIRS="src/cpp"
+    SOURCE_FILES=$(find $SOURCE_DIRS -regextype egrep -regex ".*/*\.(c|cxx|cpp|cc|h|hxx|hpp)$" ! -path '*/antlr-interface/*')
```
flomnes:
```suggestion
    SOURCE_FILES=$(find $SOURCE_DIRS -regextype egrep -regex ".*/*\.(c|cxx|cpp|cc|h|hxx|hpp)$")
```

flomnes (approved):

PR #998, opened 2025-02-06
Improve problem generation performance

JasonMarechal25:
This pull request includes several changes to the `ProblemGeneration` and `XpansionProblemsFromAntaresProvider` classes to improve code readability and performance. The most important changes involve removing unused parameters, introducing parallel execution, and making minor code improvements.

Performance improvements:

* [`src/cpp/lpnamer/problem_modifier/XpansionProblemsFromAntaresProvider.cpp`](diffhunk://#diff-6711aeeb4d261d96dc77158e3ef3fbedbbfa4edbe829f109862ba8825195f7dcL24-R47): Introduced parallel execution using `std::for_each` with `std::execution::par` to improve the performance of the `provideProblems` method.

Minor code improvements:

* [`src/cpp/lpnamer/problem_modifier/AntaresProblemToXpansionProblemTranslator.cpp`](diffhunk://#diff-1e9e4cca4224c8b4ee7484eac47719a885c2dd28cda60a94b39675bf91c5cc72L28-R28): Changed `auto` to `const auto&` for `constant` to ensure it is a reference.
* [`src/cpp/lpnamer/problem_modifier/include/antares-xpansion/lpnamer/problem_modifier/XpansionProblemsFromAntaresProvider.h`](diffhunk://#diff-c5500b6c318b35814a96d62c823fec977b96c4012cd2e1b9e64156371a15959dL15-R19): Changed the `antares_hebdo_problems` member to be a constant reference for better performance.

### DIFF
--- a/src/cpp/lpnamer/problem_modifier/XpansionProblemsFromAntaresProvider.cpp
+++ b/src/cpp/lpnamer/problem_modifier/XpansionProblemsFromAntaresProvider.cpp
@@ -24,7 +24,7 @@ std::vector<std::shared_ptr<Problem>> XpansionProblemsFromAntaresProvider::provi
 {
     std::vector<std::shared_ptr<Problem>> xpansion_problems;
     xpansion_problems.reserve(antares_hebdo_problems.weekCount());
-    std::vector<std::pair<Antares::Solver::WeeklyProblemId, Antares::Solver::WeeklyDataFromAntares>> weekly_problems_vector;
+    std::vector<std::pair<const Antares::Solver::WeeklyProblemId&, const Antares::Solver::WeeklyDataFromAntares&>> weekly_problems_vector;
     weekly_problems_vector.reserve(antares_hebdo_problems.weeklyProblems.size());
     for (const auto& [problem_id, hebdo_data] : antares_hebdo_problems.weeklyProblems) {
         weekly_problems_vector.emplace_back(problem_id, hebdo_data);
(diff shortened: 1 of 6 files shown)


src/cpp/lpnamer/problem_modifier/XpansionProblemsFromAntaresProvider.cpp
```diff
@@ -21,15 +24,30 @@ std::vector<std::shared_ptr<Problem>> XpansionProblemsFromAntaresProvider::provi
...
-        auto&& problem = AntaresProblemToXpansionProblemTranslator::translateToXpansionProblem(
-          antares_hebdo_problems,
-          problem_id.year,
-          problem_id.week,
-          solver_name,
-          solver_log_manager);
-        xpansion_problems.push_back(problem);
+        weekly_problems_vector.emplace_back(problem_id, hebdo_data);
     }
+    std::mutex mutex;
```
flomnes, line 35: You probably don't need a `mutex` here if you have a unique index for each iteration
```cpp
std::vector<Problem> out(nbElements);
std::for_each(std::execution::par, in.begin(), in.end(), [&](const auto& pair)
{
    std::size_t index = getIndex(context);
    out[index] = getProblem();
});
```
Although adding a `std::shared_ptr` to a `std::vector` shouldn't take too long

JasonMarechal25: We could compute an index from a WeeklyProblemId, but the mutex is clearer and more robust imo

tbittar (approved):

PR #1010, opened 2025-02-24
Coverage exclusion

JasonMarechal25:
Config_loader is huge and logic should be covered in dependencies or e2e


payetvin (approved):

flomnes (dismissed):
Consider using Sonar to compute coverage of Python sources

PR #1019, opened 2025-03-17
Remove 3rd party github action

JasonMarechal25:


.github/workflows/build_centos7.yml
```diff
@@ -26,16 +26,22 @@ jobs:
...
-      - name: Get changed files
-        id: changed-files
-        uses: tj-actions/changed-files@v45
-        with:
-          files: |
-            docker/centos7-system-deps
+      - name: Dockerfile has changed
+        run: |
+          if ${{ github.event_name == 'pull_request' }}; then
+            changed_files=$(git diff --name-only -r HEAD^1 HEAD | xargs)
```
flomnes, line 32: `HEAD^1` is HEAD's leftmost parent of level 1, so probably not what you want here.

Also, `fetch-depth: 0` on `actions/checkout` might interfere with the diff command.

.github/workflows/build_centos7.yml
```diff
@@ -26,16 +26,22 @@ jobs:
...
+          else
+            changed_files=$(git diff --name-only ${{ github.event.before }} ${{ github.event.after }} | xargs)
```
flomnes, line 34: I wasn't able to find anything related to `github.event.before` or `github.event.after` in the official doc

flomnes (commented):
Instead of grepping, you might want to call `git diff` precisely on the files that you want to check
```
git diff -- file revA revB
```

PR #1021, opened 2025-03-18
Resolve 1 reliability issue

flomnes:
As of now there are only 2 reliability issues reported by Sonar, see [this](https://sonarcloud.io/project/issues?impactSoftwareQualities=RELIABILITY&issueStatuses=OPEN%2CCONFIRMED&id=AntaresSimulatorTeam_antares-xpansion)

The goal of this PR is to remove one of them.

### DIFF
--- a/src/cpp/multisolver_interface/SolverCbc.cpp
+++ b/src/cpp/multisolver_interface/SolverCbc.cpp
@@ -95,7 +95,7 @@ void SolverCbc::free()
 void SolverCbc::write_prob_mps(const std::filesystem::path& filename)
 {
     const int numcols = get_ncols();
-    std::vector<char> shared_integrality(numcols);
+    std::vector<char> shared_integrality(static_cast<std::size_t>(numcols));
     char* integrality = shared_integrality.data();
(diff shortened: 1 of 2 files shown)


src/cpp/multisolver_interface/SolverCbc.cpp
```diff
@@ -95,8 +95,8 @@ void SolverCbc::free()
 void SolverCbc::write_prob_mps(const std::filesystem::path& filename)
 {
     const int numcols = get_ncols();
-    std::shared_ptr<char[]> shared_integrality(new char[numcols]);
-    char* integrality = shared_integrality.get();
+    std::vector<char> shared_integrality(static_cast<std::size_t>(numcols));
+    char* integrality = shared_integrality.data();
```
guilpier-code: do we need to define : 
```c++
char* integrality = shared_integrality.data();
```
?
Or we could use **shared_integrality** instead ?

flomnes: C APIs don't know about `std::vector<char>`

guilpier-code (approved):

PR #1027, opened 2025-03-31
[ANT-2210] Vectorisation problem generation

JasonMarechal25:
## In memory, build problems inside loop
ProblemGeneration build all problems then treat them. When there is a lot of problems to process we can hit out of memory error.
We know build problems inside the loop treating them.

## trim_in_place
Add a new function _trim_in_place_ modify the string instead of building a new one. Used to avoid allocations.

## Reduce dependency to Problem in YearlyWeightsWriter
We don't need the full problem, just the year. 

## Fix unwanted copy of WeeklyDataFromAntares
When translating a problem we access WeeklyDataFromAntares in the map but a copy was made. Since the struct is quite big we take it by reference if possible.

## Clear hebdo data after usage
After a problem is created from hebdo data they became unnecessary. We clear them in order to free up some space for subsequent problems.


conception/Architecture_decision_records/[019] Vectorise problem creation in lpnamer.md
```diff
@@ -0,0 +1,21 @@
+# Vectorise problem creation in LPNamer
+
+## Accepted [31 Mar. 2025]
+
+## Context
+
+LPNamer build all problems before processing them. When there are a lot of problems to process we can hit the limit of
+memory. This is the same kind of issue encounter during Benders.
```
Comment (tbittar) [inline conception/Architecture_decision_records/[019] Vectorise problem creation in lpnamer.md]: encountered

src/cpp/lpnamer/main/ProblemGeneration.cpp
```diff
@@ -376,78 +376,133 @@ void ProblemGeneration::RunProblemGeneration(
...
+    else // API
     {
-        ProcessWeights(problems_and_data,
-                       xpansion_output_dir,
-                       weights_file,
-                       solver_config_.Name(),
-                       logger);
+        auto mps_file_writer = std::make_shared<FileWriter>(lpDir_);
+
+        // vector of pair for parralellisation
```
tbittar: parallelization

tests/cpp/lp_namer/WeightsFileWriterTest.cpp
tbittar, line 1: The class `MockProblem` seems unused now

src/cpp/lpnamer/input_reader/WeightsFileWriter.cpp
```diff
@@ -30,20 +30,20 @@ YearlyWeightsWriter::YearlyWeightsWriter(
 }
 
 void YearlyWeightsWriter::CreateWeightFile(
-  const std::vector<std::pair<std::shared_ptr<Problem>, ProblemData>>& problems_and_data)
+  const std::vector<std::pair<int, ProblemData>>& problems_and_data)
```
tbittar: Rename arg as `year_and_data` for clarity

tbittar (approved):

PR #1029, opened 2025-03-31
Move weight processing in dedicated class

JasonMarechal25:
Move Weight processing in dedicated class WeightFileProcessor

* Reduce number of dependencies in ProblemGeneration
* Better isolate responsability


src/cpp/lpnamer/problem_modifier/WeightFileProcessor.cpp
```diff
@@ -0,0 +1,53 @@
+
+#include "antares-xpansion/lpnamer/problem_modifier/WeightFileProcessor.h"
+
+#include "antares-xpansion/lpnamer/input_reader/GeneralDataReader.h"
+#include "antares-xpansion/lpnamer/input_reader/MpsTxtWriter.h"
+#include "antares-xpansion/lpnamer/input_reader/WeightsFileReader.h"
+#include "antares-xpansion/lpnamer/input_reader/WeightsFileWriter.h"
+
+void WeightFileProcessor::ProcessWeights(
+  const std::vector<std::pair<int, ProblemData>>& problems_and_data,
```
tbittar: Rename arg as `year_and_data` for clarity

src/cpp/lpnamer/problem_modifier/WeightFileProcessor.cpp
```diff
@@ -0,0 +1,53 @@
...
+  const std::filesystem::path& weights_file,
+  const std::string& solver_name,
+  const std::shared_ptr<ProblemGenerationLog::ProblemGenerationLogger> logger)
+{
+    if (weights_file.empty())
+    {
+        return;
+    }
+    std::vector<std::pair<int, ProblemData>> year_and_data;
+    std::ranges::transform(problems_and_data,
```
tbittar, line 48: Is it possible / better to do this transform in `ProblemGeneration.cpp` so that this operation is made in the same file both in API, archive and file mode, and so that we have a unique signature for `ProcessWeights` ?

JasonMarechal25: Yes.
There is still some needed refactoring in _ProblemGeneration.cpp_ at that point it should converge

tbittar (approved):

PR #1036, opened 2025-04-14
Refactor MergeMPS

ianmnz:
Refactor as a preparation for the Pathway investment in Xpansion


src/cpp/benders/merge_mps/MergeMPS.cpp
```diff
@@ -1,269 +1,368 @@
...
+ *
+ * \param root_dir : Path to root directory where files are
+ *
+ * \param structure : Mapping of files and investments candidates
+ *
+ * \param solver_to_use : Solver name that will be used in back-end
+ *
+ * \param merged_solver : Ref to merged solver
+ */
+CouplingMap AbstractMergeMPS::get_candidates(const CouplingMap& structure)
```
tbittar: This function does much more than just `get_candidates` as it is also responsible for the effective merge of problems. Either rename it accordingly and/or refactor a bit more

src/cpp/benders/merge_mps/MergeMPS.cpp
```diff
@@ -1,269 +1,368 @@
...
+        if (filename != _options.MASTER_NAME)
         {
-            solver_l->read_prob_mps(problem_name);
-        }
-        StandardLp lpData(*solver_l);
-        std::string varPrefix_l = "prob" + std::to_string(cntProblems_l) + "_";
+            const int nb_cols{ptr_solver->get_ncols()};
 
-        lpData.append_in(mergedSolver_l, varPrefix_l);
+            IntVector indices(nb_cols);
```
tbittar: I think we are happy to progressively get rid of `IntVector` in favor of `std::vector<int>` (same for `double` and so on)

src/cpp/benders/merge_mps/MergeMPS.cpp
```diff
@@ -1,269 +1,368 @@
...
 
-    SolverFactory factory;
-    std::string solver_to_use = (_options.SOLVER_NAME == "COIN") ? "CBC" : _options.SOLVER_NAME;
-    SolverAbstract::Ptr mergedSolver_l = factory.create_solver(solver_to_use);
-    mergedSolver_l->set_output_log_level(_options.LOG_LEVEL);
+    const CouplingMap input = CouplingMapGenerator::BuildInput(structure_path,
+                                                               _logger.get(),
+                                                               "Merge mps");
+
+    const CouplingMap candidates = get_candidates(input);
```
tbittar: It unclear what `candidates` is. If I understand well:
- `input` is the `CouplingMap` read from the structure file, mapping a problem names to a map : candidates name -> id
- Here `candidates` maps candidate names to a map : filenames -> id

So for me the two objects are not of the same "type" (even though they are `std::map<std::string, std::map<std::string, int>>`). It seems that they contain the same information though. Then :
- the names should be more representative of what is inside the data structure.
- is it possible / clever to rearrange `candidates` to be the same kind of map as `input` ? I understand that it is easier to iterate over the map in this order when adding the coupling constraints. Otherwise, create a new data type ?

src/cpp/benders/merge_mps/MergeMPS.cpp
```diff
@@ -1,269 +1,368 @@
...
-    DblVector rhs(nrows, 0);
-    CharVector sense(nrows, 'E');
-    solver_addrows(*mergedSolver_l, sense, rhs, {}, mstart, cindex, values);
+    return candidates;
+}
 
+void AbstractMergeMPS::export_problem()
+{
+    const auto output_root = std::filesystem::path(_options.OUTPUTROOT);
     _logger->display_message("Problems merged.");
```
tbittar: This log may be more suited at the end of the function `get_candidates` ?

src/cpp/benders/merge_mps/MergeMPS.cpp
```diff
@@ -1,269 +1,368 @@
...
-        sol_infos.problem_status = "OPTIMAL";
-    }
-    else
+    sol_infos.solution.candidates.clear();
+    sol_infos.solution.candidates.reserve(investments.size());
+
+    for (const auto& [var_name, var_value]: investments)
     {
-        sol_infos.problem_status = "ERROR";
+        const Output::CandidateData candidate_data{var_name, var_value, -1, -1};
```
tbittar: Why don't we get the bounds of the candidates (I know it is like that in the current code) ? -> Therefore the writer outputs -1 for the bounds ? We could get them easily, does it worth it ?

src/cpp/benders/merge_mps/MergeMPS.cpp
```diff
@@ -1,269 +1,368 @@
...
     {
-        const double weight(_options.SLAVE_WEIGHT_VALUE);
-        return 1 / weight;
+        return 1.0 / _options.SLAVE_WEIGHT_VALUE;
     }
     else
     {
-        if (_options.weights.find(name) == _options.weights.end())
+        const auto found = _options.weights.find(name);
+        if (found == _options.weights.end())
```
tbittar:
```suggestion
        if (const auto found = _options.weights.find(name); found == _options.weights.end())
```

src/cpp/benders/merge_mps/MergeMPS.cpp
```diff
@@ -1,269 +1,368 @@
...
+
+/**
+ * \brief Add coupling equality constraints between subproblems
+ *
+ * \param candidates : Mapping of investment candidates
+ */
+void MergeMasterSubproblemMPS::add_coupling_constraints(const CouplingMap& candidates)
+{
+    // TODO Investigate why following check
+    // TODO creates a segfault when structure.txt is empty
```
tbittar: Is it needed ? Remove if unneccessary

src/cpp/benders/merge_mps/MergeMPS.cpp
```diff
@@ -1,269 +1,368 @@
...
+
+    _logger->display_message("About to add " + std::to_string(nb_rows_reserve)
+                             + " coupling constraints");
+
+    IntVector var_offsets;
+    IntVector var_indices;
+    DblVector var_values;
+    var_values.reserve(nb_elem_reserve);
+    var_indices.reserve(nb_elem_reserve);
+    var_offsets.reserve(nb_rows_reserve + 1);
```
tbittar: I prefer the names respectively `dmatval`, `mclind` and `mstart` to match the names used in the Xpress API and use everywhere else in the code when manipulating rows or columns in solver objects

src/cpp/benders/merge_mps/MergeMPS.cpp
```diff
@@ -1,269 +1,368 @@
...
+    // TODO creates a segfault when structure.txt is empty
+    // if (candidates.empty())
+    // {
+    //     return;
+    // }
+
+    size_t nb_elem_reserve{0}; // 2-permutation of variables
+    for (const auto& [_, file_mapping]: candidates)
+    {
+        nb_elem_reserve += file_mapping.size() * (file_mapping.size() - 1);
```
tbittar: For each variable, if we have n subproblems in the `file_mapping`, we add n-1 coupling constraints. Each of the constraints has 2 elements. Therefore, I would have only allocated `2 * (file_mappping.size() - 1)` am I wrong ?
We could simply run in debug mode and see what is the value of `nelem` at the end of this function and compare with `nb_eleme_reserve`

src/cpp/exe/merge/main.cpp
```diff
@@ -8,6 +8,7 @@
 #include "antares-xpansion/benders/factories/WriterFactories.h"
 #include "antares-xpansion/benders/logger/User.h"
 #include "antares-xpansion/benders/merge_mps/MergeMPS.h"
+#include "antares-xpansion/benders/merge_mps/StandardLp.h"
```
tbittar, line 11: Unused ?

tests/cpp/merge_mps/merge_mps_test.cpp
```diff
@@ -5,6 +5,7 @@
 #include "RandomDirGenerator.h"
 #include "InMemoryWriter.h"
 #include "antares-xpansion/benders/merge_mps/MergeMPS.h"
+#include "antares-xpansion/benders/merge_mps/StandardLp.h"
```
tbittar, line 8: Unused ?

tbittar (changes requested):

src/cpp/exe/merge/main.cpp
ianmnz: In the current code, it is used to initialize the following static variable:
`size_t StandardLp::appendCNT = 0;`

For now I kept it, but we can find options to change it eventually

tests/cpp/merge_mps/merge_mps_test.cpp
ianmnz: Same as previous comment

src/cpp/benders/merge_mps/MergeMPS.cpp
ianmnz: I agree with you, but this is the current behavior and I didn't want to change it for now
But we can test it and change it if your suspicions are right

src/cpp/benders/merge_mps/MergeMPS.cpp
ianmnz: Personally it's less clear to me but if you think its better this way, I can change it

src/cpp/benders/merge_mps/MergeMPS.cpp
ianmnz: For me we can remove it, but it's more of a reminder to better understand why it would be segfault instead of just doing nothing

src/cpp/benders/merge_mps/MergeMPS.cpp
ianmnz: I guess it wouldn't be hard to change it, but do we wish to output these bounds? Do we use it elsewhere?

src/cpp/benders/merge_mps/MergeMPS.cpp
ianmnz: No, because we need to do it after the addition of the coupling constraints

src/cpp/benders/merge_mps/MergeMPS.cpp
ianmnz: I agree with you. We need a better data struct for representing the variables and I am thinking about it. I would love to discuss it if you have any ideas :)

src/cpp/benders/merge_mps/MergeMPS.cpp
ianmnz: Thank god! I kept it because I didn't know if you wanted this way, but I will gladly change it

src/cpp/benders/merge_mps/MergeMPS.cpp
ianmnz: Indeed. I am thinking about better naming and refactoring some functions

src/cpp/benders/merge_mps/MergeMPS.cpp
```diff
@@ -294,75 +281,72 @@ double MergeMasterSubproblemMPS::get_objective_weight(const int nb_subproblems,
...
-        nb_elem_reserve += file_mapping.size() * (file_mapping.size() - 1);
+        for (const auto& [var_name, var_idx]: var_map)
+        {
+            variables[var_name].push_back(var_idx);
+        }
     }
 
-    const size_t nb_rows_reserve = nb_elem_reserve / 2; // choose-2-combination of variables
+    // Add n-1 new constraints per variable where n is
+    // the number of columns representing this variable
```
tbittar: where n is the number of problems where this varaible appears i.e. the number of  columns in the merged problems representing this variable

src/cpp/benders/merge_mps/MergeMPS.cpp
```diff
@@ -294,75 +281,72 @@ double MergeMasterSubproblemMPS::get_objective_weight(const int nb_subproblems,
...
+    dmatval.reserve(nb_elem_reserve);
+    mclind.reserve(nb_elem_reserve);
 
     int nb_rows{0};
     int nb_elem{0};
-    for (const auto& [var_name, file_map]: candidates)
+    for (const auto& [var_name, indices]: variables)
     {
-        _logger->display_message(var_name);
+        _logger->display_message(var_name + " : " + std::to_string(indices.size() - 1)
```
tbittar: Print the log after solver addrows, to ensure that the line is printed once the operation succeeds ?

src/cpp/benders/merge_mps/MergeMPS.cpp
```diff
@@ -342,11 +344,14 @@ void MergeMasterSubproblemMPS::add_coupling_constraints()
 
             ++nb_rows;
         }
+
+        logger_->display_message(var_name + " : " + std::to_string(indices.size() - 1)
+                                 + " coupling constraints added");
```
tbittar: Maybe the log here is `coupling constraints built` and you can add a `coupling constraints added` after `solver_addrows` which effectively adds the constraints to the solver

tbittar (approved):

src/cpp/benders/merge_mps/MergeMPS.cpp
JasonMarechal25: IntVector doesn't bring more value than std::vector<int>
Something like "IndexVector" would have more value but is debatable.
Name alias is especially encouraged when collections become complex like std:map<std::string, std::vector>, or map of map of...

src/cpp/benders/merge_mps/MergeMPS.cpp
JasonMarechal25: It depends on "where" you are. Here we are close to solvers and far from the business side, so it make sens to use names in the vocabulary of solvers. My comment would be different if we where at a higher level like in problem generation for example

PR #1045, opened 2025-04-22
Fix clang build

JasonMarechal25:
Fix build and warnings


JasonMarechal25:
Unfeasability after upgrading CBC

PR #1047, opened 2025-04-23
Refacto BendersFactory constructor

JasonMarechal25:
## Eliminate extra "rank" parameter
Rank can be access from world parameter

## Encapsulate dependencies
There are several "technical" dependencies, like loggers that push to a constructor with lot of parameters and thus less readable. By encapsulating them in a struct we simplify the constructor code and improve readability


src/cpp/benders/factories/BendersFactory.cpp
```diff
@@ -12,21 +12,12 @@
...
 
 BendersFactory::BendersFactory(const SimulationOptions& options,
-                               std::shared_ptr<ILogger> logger,
-                               std::shared_ptr<Output::OutputWriter> writer,
-                               std::shared_ptr<MathLoggerDriver> math_log_driver_,
-                               int rank,
-                               boost::mpi::environment* env,
                                boost::mpi::communicator* world,
-                               BendersLoggerBase& benders_loggers):
+                               Dependencies dependencies):
```
flomnes, line 16: Previously, `std::shared_ptr` were moved, now they're copied. Is this intended ?

src/cpp/benders/factories/BendersApp.cpp
```diff
@@ -80,14 +80,10 @@ int BendersApp::RunBenders()
...
-                               writer_,
-                               math_log_driver_,
-                               pworld_->rank(),
-                               penv_,
-                               pworld_,
-                               benders_loggers_);
+        BendersFactory factory(
+          options_,
+          pworld_,
+          BendersFactory::Dependencies{logger_, writer_, math_log_driver_, benders_loggers_});
```
flomnes, line 86: why 2 loggers here ?

src/cpp/exe/full_run/main.cpp
```diff
@@ -70,12 +70,12 @@ int main(int argc, char** argv)
     auto solver = options_parser.Solver();
     if (solver == "benders")
     {
-        auto benders_factory = BendersApp(options_file, env, world, SOLVER::BENDERS);
+        auto benders_factory = BendersApp(options_file, world, SOLVER::BENDERS);
```
flomnes, line 73: You could use a string to enum conversion utility here to avoid code duplication here
```cpp
std::optional<SOLVER> fromString(const std::string&);


if (auto solverEnum = fromString(solver))
{
    auto benders_factory = BendersApp(options_file, world, solverEnum);
    benders_factory.Run();
}
```

tests/cpp/multisolver_interface/CMakeLists.txt
```diff
@@ -20,10 +20,10 @@ target_link_libraries(test_multisolver PRIVATE
...
 
 if (CMAKE_CXX_COMPILER_ID STREQUAL "Clang")
-    target_compile_options(unit_tests PUBLIC -Wno-global-constructors)
-    target_compile_options(unit_tests PUBLIC -Wno-exit-time-destructors)
-    target_compile_options(unit_tests PUBLIC -Wno-weak-vtables)
-    target_compile_options(unit_tests PUBLIC -Wno-covered-switch-default)
+    target_compile_options(test_multisolver PUBLIC -Wno-global-constructors)
+    target_compile_options(test_multisolver PUBLIC -Wno-exit-time-destructors)
+    target_compile_options(test_multisolver PUBLIC -Wno-weak-vtables)
+    target_compile_options(test_multisolver PUBLIC -Wno-covered-switch-default)
```
flomnes: one single multiline `target_compile_options` here ?

src/cpp/benders/factories/BendersFactory.cpp
```diff
@@ -169,7 +159,9 @@ auto BendersFactory::PrepareForExecution(bool outer_loop) -> std::optional<Bende
     if (rank == 0)
     {
         if (Benders::StartUp startup;
-            startup.StudyAlreadyAchievedCriterion(options_, writer_.get(), logger_.get()))
+            startup.StudyAlreadyAchievedCriterion(options_,
+                                                  dependencies_.writer.get(),
+                                                  dependencies_.logger.get()))
```
flomnes, line 164: Note - `dependencies_.writer.get()` and `dependencies_.logger.get()` may become dangling pointers if their std::shared_ptr object releases them

src/cpp/benders/factories/BendersFactory.cpp
flomnes: This change makes the pointers preserved for a broader scope, since each copies keeps them alive. Not necessarily a bad thing

JasonMarechal25: std::shared_ptr where taken by value, so copied, then moved.
Now they are hold in a struct passed by value too, so copied. Meaning shared_ptr will be copied too. Unless the copy can be eliminated.
Nevertheless it shouldn't change the lifetime or life count of pointers

src/cpp/benders/factories/BendersFactory.cpp
JasonMarechal25: Yes but it should not happen. StudyAlreadyAchievedCriterion is a function, so it does not keep those pointer alive past the call point

src/cpp/benders/factories/BendersApp.cpp
JasonMarechal25: - logger_ is a application logger
- math_log_driver_: is a business logic logger (benders business is mathematical domain)
- benders_loggers_ is a meta logger storing both loggers.

benders_loggers_ is meant to broadcast messages to all underlying loggers. It is used in a couple places
logger_ is for application stuff, technical stuff, etc.
math_logger is for detailed mathematical logs, outputted in a differently file and formatted differently

src/cpp/exe/full_run/main.cpp
JasonMarechal25: https://github.com/AntaresSimulatorTeam/antares-xpansion/issues/1137

tests/cpp/multisolver_interface/CMakeLists.txt
JasonMarechal25: https://github.com/AntaresSimulatorTeam/antares-xpansion/pull/1139

flomnes (approved):

PR #1049, opened 2025-05-07
[ANT-3267] Feature/presolve

ianmnz:


ianmnz:
@tbittar @JasonMarechal25  here you will find the Draft for the presolve PR.
Please feel free to do a first review of the code
I left several TODO so we can discuss it together as well

src/cpp/exe/presolve/main.cpp
```diff
@@ -0,0 +1,179 @@
...
+    if (!factory.isXpress_available_)
+    {
+        std::cerr << "Error: Xpress not available" << std::endl;
+        std::exit(1);
+    }
+
+    XPRSPtr solver_ptr = std::static_pointer_cast<SolverXpress>(
+      factory.create_solver(options.SOLVER_NAME));
+
+    // TODO Default is 0. Should keep it ?
```
tbittar: Yes, let's comply with the settings in `options`, maybe set to one if `options.LOG_LEVEL >= 1`

src/cpp/exe/presolve/main.cpp
```diff
@@ -0,0 +1,179 @@
...
+    const int nbRows = solver.get_nrows();
+    const int nbCols = solver.get_ncols();
+
+    std::vector<int> rowmap(nbRows), colmap(nbCols);
+
+    solver.get_presolve_map(rowmap.data(), colmap.data());
+
+    for (int reduced_idx = 0; reduced_idx < colmap.size(); ++reduced_idx)
+    {
+        const int full_idx = colmap[reduced_idx];
```
tbittar: Considering that we add candidates variables at the end in problem_generation, they often (always ?) have the last indices in the full problem. So probably this is also the case in the presolved problem (this should be checked...). Therefore it may be even more profitable to iterate over `colmap` from the end to the beginning !

src/cpp/exe/presolve/main.cpp
```diff
@@ -0,0 +1,179 @@
...
+    std::unordered_map<int, int> full2reduced;
+
+    const int nbRows = solver.get_nrows();
+    const int nbCols = solver.get_ncols();
+
+    std::vector<int> rowmap(nbRows), colmap(nbCols);
+
+    solver.get_presolve_map(rowmap.data(), colmap.data());
+
+    for (int reduced_idx = 0; reduced_idx < colmap.size(); ++reduced_idx)
```
tbittar: To construct `full2reduced`, the natural algorithmic approach seems to iterate over `colind` and search the corresponding pair `(full_idx, reduced_idx)` within `colmap`. You have a "reversed" approach, iterating over `colMap`, and searching within `colind`. Is it for efficiency purpose ? (As `colind` has size nb candidates and `colmap` has size nb of presolved variables`) If so, maybe add a comment to explain this

src/cpp/exe/presolve/main.cpp
```diff
@@ -0,0 +1,179 @@
...
+    XPRSPtr solver_ptr = std::static_pointer_cast<SolverXpress>(
+      factory.create_solver(options.SOLVER_NAME));
+
+    // TODO Default is 0. Should keep it ?
+    // solver.set_output_log_level(1);
+
+    return solver_ptr;
+}
+
+std::unordered_map<int, int> get_presolve_map(SolverXpress& solver, const std::vector<int>& colind)
```
tbittar:
```suggestion
std::unordered_map<int, int> get_candidates_presolve_map(SolverXpress& solver, const std::vector<int>& candidatesId)
```

src/cpp/exe/presolve/main.cpp
```diff
@@ -0,0 +1,179 @@
...
+    {
+        if (filename == options.MASTER_NAME) [[unlikely]]
+        {
+            // Keep master indices untouched
+            // and only try to reduce the subproblems
+            continue;
+        }
+
+        // Get all candidate indices in a sorted array
+        std::vector<int> colind(var_map.size());
```
tbittar:
```suggestion
        std::vector<int> candidatesId(var_map.size());
```

src/cpp/exe/presolve/main.cpp
```diff
@@ -0,0 +1,179 @@
...
+        std::transform(var_map.cbegin(),
+                       var_map.cend(),
+                       colind.begin(),
+                       [](const auto pair) { return pair.second; });
+        std::sort(colind.begin(), colind.end());
+
+        // Read full problem
+        const std::filesystem::path subproblem_path = input_root_dir / filename;
```
tbittar: TODO (Not in this PR, for later on @JasonMarechal25 ) : Handle the case where problems are passed from problem_generation to presolve in memory : Need to think about the whole in-memory workflow

src/cpp/exe/presolve/main.cpp
```diff
@@ -0,0 +1,179 @@
...
+        for (const auto& [var_name, idx]: var_map)
+        {
+            reduced_couplings.at(filename).at(var_name) = full2reduced.at(idx);
+        }
+    }
+
+    return reduced_couplings;
+}
+
+void write_reduced_problems(const CouplingMap& reduced_couplings,
```
tbittar:
```suggestion
void write_reduced_couplings(const CouplingMap& reduced_couplings,
```
`write_reduced_problems` is confusing, it looks like you are writing the mps files

src/cpp/exe/presolve/main.cpp
```diff
@@ -0,0 +1,179 @@
...
+    BaseOptions options{SimulationOptions(argv[1]).get_base_options()};
+    Logger logger = std::make_shared<xpansion::logger::User>(std::cout);
+
+    logger->display_message("Starting presolve");
+
+    XPRSPtr solver_ptr = init_solver(options, logger);
+
+    const CouplingMap reduced_couplings = reduce_problems(*solver_ptr, options, logger);
+
+    write_reduced_problems(reduced_couplings, options, logger);
```
tbittar: I do not understand why the writing part of the reduced structure is split from `reduce_problems` whereas the writing part for problems themselves is done within `reduce_problems`. Maybe do a consistant choice for all writing operations (especially if later on all this could be handled in-memory or not) ?

src/cpp/exe/presolve/main.cpp
tbittar, line 1: All this is adherent to Xpress low-level interface. For later on (not in this PR), think about if/how we need to handle this with the multisolver interface

src/cpp/multisolver_interface/environment.cc
```diff
@@ -145,6 +146,12 @@ std::function<int(
   void(XPRS_CC* f_message)(XPRSprob cbprob, void* cbdata, const char* msg, int msglen, int msgtype),
   void* p)>
   XPRSsetcbmessage = nullptr;
+std::function<int(
+  XPRSprob prob,
+  void(XPRS_CC* f_message)(XPRSprob cbprob, void* cbdata, const char* msg, int msglen, int msgtype),
+  void* p,
+  int priority)>
+  XPRSaddcbmessage = nullptr;
```
tbittar: Do we still use this ? (Used for debug in the POC if I remember well)

src/cpp/multisolver_interface/environment.cc
```diff
@@ -209,6 +217,7 @@ bool XpressLoader::LoadXpressFunctions(Solver::DynamicLibrary* xpress_dynamic_li
     xpress_dynamic_library->GetFunction(&XPRSgetdblattrib, "XPRSgetdblattrib");
     xpress_dynamic_library->GetFunction(&XPRSgetmipsol, "XPRSgetmipsol");
     xpress_dynamic_library->GetFunction(&XPRSsetcbmessage, "XPRSsetcbmessage");
+    xpress_dynamic_library->GetFunction(&XPRSaddcbmessage, "XPRSaddcbmessage");
```
tbittar: Same question

src/cpp/multisolver_interface/include/antares-xpansion/multisolver_interface/environment.h
```diff
@@ -480,6 +481,7 @@ extern std::function<int(XPRSprob prob, int attrib, double* p_value)> XPRSgetdbl
 extern std::function<int(XPRSprob prob, double x[], double slack[], double duals[], double djs[])> XPRSgetlpsol;
 extern std::function<int(XPRSprob prob, double x[], double slack[])> XPRSgetmipsol;
 extern std::function<int(XPRSprob prob, void (XPRS_CC *f_message)(XPRSprob cbprob, void* cbdata, const char* msg, int msglen, int msgtype), void* p)> XPRSsetcbmessage;
+extern std::function<int(XPRSprob prob, void (XPRS_CC *f_message)(XPRSprob cbprob, void* cbdata, const char* msg, int msglen, int msgtype), void* p, int priority)> XPRSaddcbmessage;
```
tbittar: May be removed ?

src/python/antares_xpansion/driver.py
```diff
@@ -51,10 +51,12 @@ def __init__(self, config_loader: ConfigLoader):
             SolversExe(
                 self.config_loader.benders_exe(),
                 self.config_loader.merge_mps_exe(),
+                self.config_loader.presolve_exe(),
                 self.config_loader.outer_loop_exe()),
             self.config_loader.options_file_name(),
             self.config_loader.mpi_exe(),
         )
+        # TODO options.json is empty at this stage. How to force the solver to be Xpress?
```
tbittar: I think we need to comply with the user's choice in the settings.ini. Maybe notice early from the Python orchestrator if the presolve part is called without Xpress, and raise an exception in this case ?

src/python/antares_xpansion/input_parser.py
```diff
@@ -38,7 +38,7 @@ def _initialize_parser(self):
             "--method",
             dest=LauncherOptionsKeys.method_key(),
             type=str,
-            choices=["benders", "mergeMPS", "adequacy_criterion"],
+            choices=["benders", "mergeMPS", "presolve", "adequacy_criterion"],
```
tbittar: I am not sure that `presolve` has the same status as other choices in this list. For me, it looks more like a step (such as `problem_generation`), that could be optional, called only if solver Xpress is chosen, and that should appear (as an optional step) in the full run. Any remark on these reflexions are welcome @JasonMarechal25 @ianmnz 
Moreover, if included in the full run workflow, there will be a work to handle outputs of the simulation (as operational costs would be those of the presolved pb, not the original one. Candidates should be ok, and that is the most important part).

src/cpp/exe/presolve/main.cpp
ianmnz: Interesting. I will test it in the reversed order then

src/cpp/exe/presolve/main.cpp
ianmnz: To be honest, for no hard reason. Just to split into the different logical steps I saw ("solver initialization", "reduced coupling generation" and "output results"), but indeed it can go inside reduce_problems as the others :)

src/cpp/exe/presolve/main.cpp
ianmnz: Initially I used the Xpress low-level API, but it had too much duplicate code, the management of error status was verbose and the destructor needed to be carefully written. All of it was pretty well done in SolverXpress directly so I chose to use it and enrich it for our specific needs

src/cpp/multisolver_interface/environment.cc
ianmnz: Initially I used it before switching to SolverXpress interface and I forgot to delete it. Good catch!

src/python/antares_xpansion/driver.py
ianmnz: Hm... you are suggesting that it's the job of the Python orchestrator to verify the compliance of the solver? I can look how to do it

src/python/antares_xpansion/input_parser.py
ianmnz: It's true that, given the current setting, presolve is seen more as an solver option itself than a step before the solver. We would need to think on how to integrate in a full run, for instance

src/cpp/exe/presolve/main.cpp
ianmnz: What I thought was that, since `colind` is pretty small compared to `colmap`, (mostly) sorting it and using a binary search would be faster than doing the opposite.
Here is my reasoning: Being `#colmap = N >> m = #colind`
sort `O(mlogm) << O(NlogN)`
search `O(Nlogm) > O(mlogN)`
together `O(Nlogm) < O(NlogN)`

src/cpp/exe/presolve/main.cpp
ianmnz: From the Xpress page, the presolve can add cols and rows to a problem. Still, I don't think it will have the same order of magnitude than a weekly problem. I think it is a good option to reverse it

ianmnz:
@tbittar @JasonMarechal25 I left some TODO in the code. I don't think I will have the time to fix them, but we can discuss it this week so I can explain them to you. Most of them are small doubts I had related to the overall design and should be easy to address

ianmnz:
The last TODO is the big one. I left quite a comment in the code (@tbittar knows about it already). I will leave here some references I found that could help you when fixing it:
 - [Optimizer Error Code 268](https://www.fico.com/fico-xpress-optimization/docs/latest/solver/optimizer/HTML/chapter9.html?scroll=section9001) : This is the code error I got when I tried to run Benders. It is thrown when we call 'chg_obj' in the 'SubproblemWorker' constructor;
 - [PRESOLVESTATE attr code](https://www.fico.com/fico-xpress-optimization/docs/latest/solver/optimizer/HTML/javadoc/constant-values.html) Attribute used to check if the problem is in a presolved state. It uses a bitmap with the PRESOLVESTATE_XXX [attributes](https://www.fico.com/fico-xpress-optimization/docs/latest/solver/optimizer/HTML/PRESOLVESTATE.html);
 - [PresolveAPI](https://www.fico.com/fico-xpress-optimization/docs/latest/solver/optimizer/HTML/chapterTopics.html?scroll=secoptimizertopics-Presolve)
 - [Working with presolve](https://www.fico.com/fico-xpress-optimization/docs/dms2019-03/solver/optimizer/HTML/chapter5_sec_section5001.html)
 - [Presolve question in Forum](https://fico.my.site.com/FICOCommunity/s/question/0D54w0000APch2eCQB/is-it-possible-to-export-a-presolved-model)
 - [Postsolve](https://www.fico.com/fico-xpress-optimization/docs/latest/solver/optimizer/HTML/XPRSpostsolve.html)

tbittar:
> The last TODO is the big one. I left quite a comment in the code (@tbittar knows about it already). I will leave here some references I found that could help you when fixing it:
> 
>     * [Optimizer Error Code 268](https://www.fico.com/fico-xpress-optimization/docs/latest/solver/optimizer/HTML/chapter9.html?scroll=section9001) : This is the code error I got when I tried to run Benders. It is thrown when we call 'chg_obj' in the 'SubproblemWorker' constructor;
> 
>     * [PRESOLVESTATE attr code](https://www.fico.com/fico-xpress-optimization/docs/latest/solver/optimizer/HTML/javadoc/constant-values.html) Attribute used to check if the problem is in a presolved state. It uses a bitmap with the PRESOLVESTATE_XXX [attributes](https://www.fico.com/fico-xpress-optimization/docs/latest/solver/optimizer/HTML/PRESOLVESTATE.html);
> 
>     * [PresolveAPI](https://www.fico.com/fico-xpress-optimization/docs/latest/solver/optimizer/HTML/chapterTopics.html?scroll=secoptimizertopics-Presolve)
> 
>     * [Working with presolve](https://www.fico.com/fico-xpress-optimization/docs/dms2019-03/solver/optimizer/HTML/chapter5_sec_section5001.html)
> 
>     * [Presolve question in Forum](https://fico.my.site.com/FICOCommunity/s/question/0D54w0000APch2eCQB/is-it-possible-to-export-a-presolved-model)
> 
>     * [Postsolve](https://www.fico.com/fico-xpress-optimization/docs/latest/solver/optimizer/HTML/XPRSpostsolve.html)

@JasonMarechal25 I managed to perform the presolve step with svf files by copying the matrix data to a new solver object !! Could you review the changes made in this part, espacially focusing on avoiding unecessary copies and so on (I changed the return types of some functions used by Ian so i don't know if this is the most efficient)

tbittar:
Still to do before merging this PR:

- [ ] Look at the TODOs left by Ian in the code (mostly related to the orchestration in the Python, but also some in `src/cpp/exe/presolve.cpp` and other files)
- [ ] Add some unit test for the presolve part in C++ (I will try to give you the inputs and outputs of such test)
- [ ] Add tests for the Python orchestration involving presolve
- [ ] (Maybe not in this PR, just to keep in mind) There might be needs to think about the outputs (as operational cost will not be the "true" operational cost

JasonMarechal25:
E2E tests  have been implemented but require a way to open svf/mps.
To validate presolve we run lpnamer then presolve, with xpress. It produce svf by default, we need to configure tests to use mps format.

JasonMarechal25:
Need to remove trajectory launcher and fuse with xpansion launcher

JasonMarechal25 (approved):

PR #1073, opened 2025-08-01
Complete chg_rhs function

h-fournier:
If the RHS is fixed then modify both lower and upper bounds

### DIFF
--- a/src/cpp/multisolver_interface/COIN_common_functions.cpp
+++ b/src/cpp/multisolver_interface/COIN_common_functions.cpp
@@ -94,9 +94,16 @@ void fill_rhs_from_bounds(const double* rowLower,
         {
             if (rowUpper[i] < COIN_DBL_MAX)
             {
-                auto error = LOGLOCATION + "ERROR : Row " + std::to_string(i)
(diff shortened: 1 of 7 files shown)


h-fournier:
I will add a unit-test before merging

src/cpp/multisolver_interface/COIN_common_functions.cpp
```diff
@@ -94,9 +94,16 @@ void fill_rhs_from_bounds(const double* rowLower,
         {
             if (rowUpper[i] < COIN_DBL_MAX)
             {
-                auto error = LOGLOCATION + "ERROR : Row " + std::to_string(i)
-                             + " has two different RHS, both right and left.";
-                throw GenericSolverException(error);
+                if (rowLower[i] == rowUpper[i])
```
tbittar: I do not really get when we can get into this `if` block without being passed into the one at line 89

h-fournier: True, I removed this modification

tbittar (approved):

PR #1079, opened 2025-08-18
Refacto MasterGeneration

JasonMarechal25:
Split structure generation from master generation


src/cpp/lpnamer/problem_modifier/StructureGeneration.cpp
```diff
@@ -0,0 +1,74 @@
...
+{
+}
+
+void StructureGeneration::operator()(const std::vector<Candidate>& candidates,
+                                     const Couplings& couplings) const
+{
+    write_structure_file(candidates, couplings);
+}
+
+void StructureGeneration::write_structure_file(const std::vector<Candidate>& candidates,
```
flomnes, line 45: `StructureGeneration::write_structure_file` and `StructureGeneration::operator` have the same arguments, while `StructureGeneration::operator` just calls `StructureGeneration::write_structure_file`.

In my opinion we can just keep `StructureGeneration::write_structure_file` which looks more clear (the caller knows it will write a file).

src/cpp/lpnamer/main/ProblemGeneration.cpp
```diff
@@ -425,16 +426,17 @@ void ProblemGeneration::RunProblemGeneration(
...
-                                       file_writer,
-                                       configuration_manager_.Format());
+    auto&& candidates = MasterGeneration(xpansion_output_dir,
+                                         solver_config_.Name(),
+                                         logger,
+                                         solver_log_manager,
+                                         file_writer,
+                                         configuration_manager_.Format())(links,
+                                                                          master_formulation,
+                                                                          additionalConstraints);
```
flomnes: Split in two for clarity
```suggestion
MasterGeneration environment(xpansion_output_dir,
                             solver_config_.Name(),
                             logger,
                             solver_log_manager,
                             file_writer,
                             configuration_manager_.Format());
auto&& candidates = environment(links, master_formulation, additionalConstraints);
```

src/cpp/lpnamer/problem_modifier/StructureGeneration.cpp
JasonMarechal25: Indeed.
I've rename the operator as "generate" method for clarity

flomnes (approved):

PR #1086, opened 2025-09-05
Remove shared_ptr alias

JasonMarechal25:
The alias tended to make us use wrong patterns with shared_ptr, passing shared_ptr& or const shared_ptr& instead of plain shared_ptr

### DIFF
--- a/tests/cpp/solvers_interface/test_exceptions.cpp
+++ b/tests/cpp/solvers_interface/test_exceptions.cpp
@@ -2,157 +2,211 @@
 #include "antares-xpansion/xpansion_interfaces/LogUtils.h"
 #include "catch2.hpp"
 
-TEST_CASE("InvalidStatusException", "[exceptions][invalid_status]") {
-  SolverFactory factory;
-  for (auto const& solver_name : factory.get_solvers_list()) {
+TEST_CASE("InvalidStatusException", "[exceptions][invalid_status]")
+{
+    SolverFactory factory;
+    for (const auto& solver_name: factory.get_solvers_list())
+    {
+        //========================================================================================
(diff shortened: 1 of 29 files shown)


tests/cpp/solvers_interface/test_exceptions.cpp
```diff
@@ -2,157 +2,211 @@
...
+        std::shared_ptr<SolverAbstract> solver = factory.create_solver(solver_name);
+
+        try
+        {
+            solver->set_algorithm("BARRIER");
+        }
+        catch (InvalidSolverOptionException& ex)
+        {
+            REQUIRE(ex.ErrorMessage() == "Invalid option 'set_algorithm : BARRIER' for solver.");
+        }
```
flomnes, line 157: Note this code doesn't actually check that an exception is thrown. It only checks that if an exception is thrown, the error message is as expected.

tests/cpp/solvers_interface/test_reading_problem.cpp
```diff
@@ -658,7 +658,7 @@ TEST_CASE("Testing copy constructor", "[init][copy-constructor]")
 
             //========================================================================================
             // Declare copy prob
-            SolverAbstract::Ptr solver2 = factory.copy_solver(*solver);
+            std::shared_ptr<SolverAbstract> solver2 = factory.copy_solver(*solver);
```
flomnes, line 661:
```suggestion
            auto solver2 = factory.copy_solver(*solver);
```

flomnes (approved):
A lot of the `std::shared_ptr<SolverAbstract>` can be replaced by `auto`, depending on style preferences.

PR #1087, opened 2025-09-05
Use atomic for multi thread protection

JasonMarechal25:
Change static variable at global scope for atomic static variable at function scope.
* Prevent static initialization fiasco
* Protect multithreading access

### DIFF
--- a/src/cpp/multisolver_interface/SolverCbc.cpp
+++ b/src/cpp/multisolver_interface/SolverCbc.cpp
@@ -55,7 +55,7 @@ SolverCbc::SolverCbc(const SolverAbstract& toCopy):
     else
     {
         number_of_problems_counter() -= 1;
-        throw InvalidSolverForCopyException(toCopy->get_solver_name(), name_, LOGLOCATION);
+        throw InvalidSolverForCopyException(toCopy.get_solver_name(), name_, LOGLOCATION);
     }
 }
 
--- a/src/cpp/multisolver_interface/SolverCbc.cpp
+++ b/src/cpp/multisolver_interface/SolverCbc.cpp
@@ -7,7 +7,7 @@ namespace
 {
 std::atomic<int>& number_of_problems_counter()
 {
-    static std::atomic<int> counter;
+    static std::atomic<int> counter{0};
     return counter;
 }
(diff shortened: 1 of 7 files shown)


src/cpp/multisolver_interface/include/antares-xpansion/multisolver_interface/SolverAbstract.h
```diff
@@ -6,6 +6,8 @@
 #include <iostream>
 #include <list>
 #include <memory>
+#include <mutex>
+#include <shared_mutex>
```
flomnes:
```suggestion
```

src/cpp/multisolver_interface/SolverCbc.cpp
```diff
@@ -44,20 +54,20 @@ SolverCbc::SolverCbc(const SolverAbstract& toCopy):
...
         throw InvalidSolverForCopyException(toCopy.get_solver_name(), name_, LOGLOCATION);
     }
 }
 
 SolverCbc::~SolverCbc()
 {
-    _NumberOfProblems -= 1;
-    SolverCbc::free();
+    number_of_problems_counter() -= 1;
+    free();
```
flomnes, line 65:
```suggestion
    SolverCbc::free();
```

flomnes (approved):

src/cpp/multisolver_interface/SolverCbc.cpp
```diff
@@ -1,13 +1,23 @@
...
 #include "COIN_common_functions.h"
 using namespace std::literals;
 
+namespace
+{
+std::atomic<int>& number_of_problems_counter()
+{
+    static std::atomic<int> counter{0};
+    return counter;
+}
```
flomnes, line 14:
```suggestion
std::atomic<int> number_of_problems_counter{0};
```

flomnes: same same ?

flomnes: See https://en.cppreference.com/w/cpp/language/siof.html

PR #1088, opened 2025-09-05
Cleanup solver copy constructor

JasonMarechal25:
Delegate dynamic cast to factory
Use standard pattern for copy constructor

### DIFF
--- a/src/cpp/multisolver_interface/SolverClp.cpp
+++ b/src/cpp/multisolver_interface/SolverClp.cpp
@@ -1,13 +1,23 @@
 #include "SolverClp.h"
 
+#include <atomic>
+
 #include "COIN_common_functions.h"
 using namespace std::literals;
+
+namespace
+{
+std::atomic<int>& number_of_problems_counter()
+{
+    static std::atomic<int> counter{0};
+    return counter;
+}
+} // namespace
 
 /*************************************************************************************************
 -----------------------------------    Constructor/Desctructor
 --------------------------------
 *************************************************************************************************/
-int SolverClp::_NumberOfProblems = 0;
 
 SolverClp::SolverClp(const SolverLogManager& log_manager):
     SolverClp()
@@ -21,7 +31,7 @@ SolverClp::SolverClp(const SolverLogManager& log_manager):
 
 SolverClp::SolverClp()
 {
-    _NumberOfProblems += 1;
+    number_of_problems_counter() += 1;
     set_output_log_level(0);
 }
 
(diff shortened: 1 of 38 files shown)


src/cpp/multisolver_interface/SolverFactory.cpp
```diff
@@ -240,6 +240,14 @@ std::shared_ptr<SolverAbstract> SolverFactory::create_solver(
     return create_solver(solver_config, log_manager);
 }
 
+template<class SolverT>
+std::shared_ptr<SolverT> copy(const SolverAbstract& to_copy)
+{
+    std::string solver_name = to_copy.get_solver_name();
```
flomnes: unused local variable
```suggestion
```

src/cpp/multisolver_interface/SolverClp.cpp
```diff
@@ -35,23 +35,14 @@ SolverClp::SolverClp()
...
-        }
-    }
-    else
+    _clp = ClpSimplex(toCopy._clp);
+    _fp = toCopy._fp;
+    if (_fp)
     {
-        number_of_problems_counter() -= 1;
-        throw InvalidSolverForCopyException(toCopy.get_solver_name(), name_, LOGLOCATION);
+        _clp.messageHandler()->setFilePointer(toCopy._fp);
```
flomnes: No difference, but more clear in my opinion (set the "copied file pointer" to the "copied _clp")
```suggestion
        _clp.messageHandler()->setFilePointer(_fp);
```

src/cpp/multisolver_interface/SolverFactory.cpp
```diff
@@ -240,26 +240,33 @@ std::shared_ptr<SolverAbstract> SolverFactory::create_solver(
...
-    std::transform(solver_name.begin(), solver_name.end(), solver_name.begin(), ::toupper);
+    std::ranges::transform(solver_name, solver_name.begin(), ::toupper);
     if (solver_name.empty())
     {
         throw InvalidSolverNameException(solver_name, LOGLOCATION);
     }
     if (isXpress_available_ && solver_name == XPRESS_STR)
     {
-        return std::make_shared<SolverXpress>(to_copy);
+        return copy<SolverXpress>(to_copy);
```
flomnes: You could have a `SolverAbstract* clone() const` virtual method here, so this code would look like
```cpp
return std::shared_ptr<SolverAbstract>(to_copy.clone());
```
and you wouldn't rely on the `get_solver_name()` method. In fact, you wouldn't need `SolverFactory::copy_solver` anymore.

src/cpp/multisolver_interface/SolverFactory.cpp
```diff
@@ -8,6 +8,8 @@
 #include "SolverCbc.h"
 #include "SolverClp.h"
 #endif
+#include <variant>
+
```
flomnes:
```suggestion
```

flomnes (approved):
`isXpress_available_` is no longer checked but we can assume that if we already have a `SolverXpress` instance to copy, then Xpress is available.

PR #1089, opened 2025-09-05
Keep Xpress context alive

JasonMarechal25:
Instead of freeing Xpress context every time the number of solver reach 0, we keep the context alive for the program duration

### DIFF
--- a/src/cpp/multisolver_interface/SolverFactory.cpp
+++ b/src/cpp/multisolver_interface/SolverFactory.cpp
@@ -29,6 +29,7 @@ void GetAvailableSolversInternal(std::shared_ptr<ILoggerXpansion> logger)
 {
     if (available_solvers.empty())
     {
+        static XpressManager xpress_manager;
         LoadXpress::XpressLoader xpress_loader(std::move(logger));
         if (xpress_loader.XpressIsCorrectlyInstalled(true))
         {
--- a/src/cpp/multisolver_interface/SolverXpress.cpp
+++ b/src/cpp/multisolver_interface/SolverXpress.cpp
@@ -14,7 +14,6 @@ using namespace std::literals;
 -----------------------------------    Constructor/Desctructor
 --------------------------------
 *************************************************************************************************/
-std::mutex SolverXpress::license_guard;
 const std::map<int, std::string> TYPETONAME = {{1, "rows"}, {2, "columns"}};
 
 namespace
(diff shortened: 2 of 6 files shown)


src/cpp/multisolver_interface/environment.cc
```diff
@@ -564,10 +564,6 @@ bool XpressLoader::initXpressEnv(bool verbose, int xpress_oem_license_key)
 bool XpressLoader::XpressIsCorrectlyInstalled(bool verbose)
 {
     bool correctlyInstalled = initXpressEnv(verbose);
-    if (correctlyInstalled)
-    {
-        XPRSfree();
-    }
     return correctlyInstalled;
```
flomnes, line 567: Method `XpressIsCorrectlyInstalled` becomes redundant with `initXpressEnv` now

src/cpp/multisolver_interface/SolverXpress.cpp
```diff
@@ -37,26 +36,51 @@ SolverXpress::SolverXpress(const SolverLogManager& log_manager):
...
+    static std::mutex license_guard;
+    return license_guard;
+}
+} // namespace
+
+XpressManager::XpressManager()
+{
+    std::lock_guard<std::mutex> guard(instance_guard());
+    _loader.initXpressEnv();
+    if (_loader.XpressIsCorrectlyInstalled())
```
flomnes: `_loader.initXpressEnv` is called twice it seems

src/cpp/multisolver_interface/SolverXpress.cpp
```diff
@@ -37,26 +36,51 @@ SolverXpress::SolverXpress(const SolverLogManager& log_manager):
...
-        zero_status_check(status, "initialize XPRESS environment", LOGLOCATION);
+        int status = XPRSinit(nullptr);
+        SolverAbstract::zero_status_check(status, "initialize XPRESS environment", LOGLOCATION);
     }
+}
 
-    number_of_problems_counter() += 1;
+XpressManager::~XpressManager()
+{
+    std::lock_guard<std::mutex> guard(instance_guard());
```
flomnes: C++17 has [CTAD](https://en.cppreference.com/w/cpp/language/class_template_argument_deduction.html)
```suggestion
    std::lock_guard guard(instance_guard());
```

src/cpp/multisolver_interface/SolverFactory.cpp
```diff
@@ -29,6 +29,7 @@ void GetAvailableSolversInternal(std::shared_ptr<ILoggerXpansion> logger)
 {
     if (available_solvers.empty())
     {
+        static XpressManager xpress_manager;
```
flomnes, line 32: `XpressManager` has a `XpressLoader` member, but it looks like it's not used here

flomnes (commented):
You now have an `XpressLoader`, `XpressManager`, `SolverXpress`, it's getting a bit complex. I'm not sure this complexity is really needed to achieve what you want

Consider using [`std::call_once`](http://en.cppreference.com/w/cpp/thread/call_once.html).

src/cpp/multisolver_interface/SolverFactory.cpp
JasonMarechal25: Yes. XpressManager handle RAII XpressInit/XpressFree.
XpressLoader is used to know if Xpress is available

We could "hide" xpressLoader and have XpressManager as a facade providing bool:XpressIsCorrectlyInstalled
But it would not be simple to do so.
XpressManager is a static instance to be init once and destroyed once at the end of the program lifetime. We can't inject a logger in it that we want to have in XpressLoader for logging purpose.

flomnes (approved):

PR #1119, opened 2025-11-24
feat: update generalData.ini of each trajectory study

JasonMarechal25:


src/python/antares_xpansion/trajectory/driver_multiple_problem_gen.py
```diff
@@ -180,15 +182,58 @@ def _launch_executable(self):
...
+        """
+        Update general data settings for a single study
+        Similar to XpansionDriver.update_study_settings
+        """
+        settings_dir = os.path.normpath(
+            os.path.join(study_path, self.settings)
+        )
+        # For trajectory mode, we assume is_accurate=False (Economy mode)
+        # This can be made configurable if needed
+        gen_data_proc = GeneralDataProcessor(settings_dir, is_accurate=False)
```
tbittar: We should already read the accurate/fast mode of each yearly study from now on, otherwise we may for instance have different (and unexpected/unwanted) behavior between running a single year study with "classical" Xpansion and Xpansion in trajectory over one year.

tbittar (changes requested):

src/python/antares_xpansion/trajectory/driver_multiple_problem_gen.py
JasonMarechal25: Thank you.
Fixed in 70098ffaeca48770453cd99b9288408ed5616692

src/python/antares_xpansion/trajectory/driver_multiple_problem_gen.py
```diff
@@ -61,10 +62,13 @@ def __init__(self, data: MultipleProblemGenerationData):
         self.node_to_studies: Dict[str, Path] = {}
         self.node_to_weights_file: Dict[str, Path] = {}
         self.node_to_additional_constraints: Dict[str, Path] = {}
+        # Nouveau: dictionnaire mode accurate par nœud
```
tbittar, line 65: New: Comments in French haha

tbittar (approved):

PR #1125, opened 2025-11-28
Bump tbb@2022.2.0

flomnes:


JasonMarechal25 (changes requested):
TBB 2022 use oneTbbAPI. libstdc++10 is compliant with legacy tbb API and there are build error using gcc10
See this thread for more information: https://stackoverflow.com/a/67924408

In addition to this change it would be good to update requirements in documentation (developer guide) and maybe also write an ADR on the subject.

flomnes:
> TBB 2022 use oneTbbAPI. libstdc++10 is compliant with legacy tbb API and there are build error using gcc10 See this thread for more information: https://stackoverflow.com/a/67924408
> 
> In addition to this change it would be good to update requirements in documentation (developer guide) and maybe also write an ADR on the subject.

In the CentOS 7, OracleLinux 8 and Ubuntu 22.04 pipelines, g++-11 is used, so there should not be any issues

Doc can be written

JasonMarechal25 (approved):

JasonMarechal25:
Tested locally on ubuntu 22.04. Parallelism is preserved
Tested on production: it runs

PR #1132, opened 2025-12-04
fix: [ANT-4229] Windows exe path

JasonMarechal25:
merge_weights_trajectory executable is not found on windows. Driver look into *merge_weights_trajecotry* when *merge_weights_trajecotry.exe* exists


src/python/antares_xpansion/trajectory/trajectory_config.py
```diff
@@ -92,7 +92,16 @@ def get_executable_path(self, exe_name: str):
...
+        fullpath = (Path(self.install_dir) / exe_name).resolve()
+        if not fullpath.is_file():
+            newpath = fullpath.with_suffix('exe')
+            if newpath.is_file():
+                return newpath
+            else:
+                raise FileNotFoundError(
+                    f"Executable {exe_name} not found in install dir {self.install_dir}"
+                )
+        return fullpath
```
flomnes:
```suggestion
    assert hasattr(self, "install_dir"), "install_dir attribute is missing"

    install_path = Path(self.install_dir).resolve()
    print(f"Looking for executable '{exe_name}' in: {install_path}")

    candidate = install_path / exe_name

    # Direct match
    if candidate.is_file():
        return candidate

    # On Windows, try adding .exe suffix automatically
    if os.name == "nt":
        candidate_exe = candidate.with_suffix(".exe")
        if candidate_exe.is_file():
            return candidate_exe

    raise FileNotFoundError(f"Executable '{exe_name}' not found in {install_path}")
```

flomnes (commented):
Small refactor, apply on Windows only

payetvin (approved):

PR #1139, opened 2025-12-08
refactor: cleaner handling of clang options

JasonMarechal25:
- Use a single call to set multiple options
- Factorize in a single place the different options used


flomnes (approved):
Having a single CMake function is nice. I wonder why we need to suppress clang's warnings though

PR #1144, opened 2025-12-16
Feature/benders micro iteration

hedibouchehda1997:


src/cpp/benders/factories/BendersFactory.cpp
```diff
@@ -159,9 +163,18 @@ auto BendersFactory::PrepareForExecution(bool outer_loop) -> std::optional<Bende
     BendersBaseOptions benders_options(options_.get_benders_options());
     benders_options.EXTERNAL_LOOP_OPTIONS.DO_OUTER_LOOP = outer_loop;
 
+    std::cout<<"get in PrepareForExecution"<<std::endl;
```
JasonMarechal25: To remove or use logger

PR #1152, opened 2025-12-22
same subs in each cut whatever the nproc is

hedibouchehda1997:
Before this dev , the subproblem considered in each cut can be different with the number of process we consider when we run benders which lead to different number of iterations of the master problem (dependending on the chosen nproc).
This dev will make the cut robust and make sure that we always consider the same subproblems at each cut


docs/user-guide/get-started/options-definition.md
```diff
@@ -0,0 +1,33 @@
+# Setting of launching benders exec 
+
+It is possible to launch the benders solver directly. Indeed it's we have an executable called "benders". This executable takes as input a json file called "options.json". This file contains the follwing attirubutes :
```
a-zakir: you need to re-write this

docs/user-guide/get-started/get-started.rst
tbittar, line 1: Move this to the PR on doc

vcpkg
tbittar, line 1: Revert this change as it causes CentOS CI to crash

tbittar (changes requested):

PR #1159, opened 2026-01-05
Do not erase subproblem trace

tbittar:


tbittar:
Close as fixed by #1160

PR #1169, opened 2026-01-15
Fix invest cost with all master variables

Juliette-Gerbaux:


src/cpp/benders/benders_core/BendersBase.cpp
```diff
@@ -420,15 +420,8 @@ void BendersBase::ComputeXCut()
...
-    _data.invest_cost = 0;
-
-    std::vector<double> obj(MasterObjectiveFunctionCoeffs());
-
-    for (const auto& [col_name, value]: _data.x_cut)
-    {
-        int col_id = _master->_name_to_id[col_name];
-        _data.invest_cost += obj[col_id] * _data.x_cut[col_name];
-    }
+    // The master cost is calculated as the sum of the investment cost and the overall subproblem cost under approximation. Calculating the investment cost this way is more robust than using x_cut, as it ensures that all master variables are accounted for.
```
tbittar:
```suggestion
    // The master cost is calculated as the sum of the investment cost and the overall subproblem cost under approximation. Calculating the investment cost this way is more robust than computing c*x_cut, as it ensures that all master variables are accounted for. Indeed x_cut only contain variables that couple the master and the subproblems and not master only variables (which may have an associated cost).
```

src/cpp/benders/benders_core/BendersBase.cpp
```diff
@@ -1422,13 +1415,8 @@ bool BendersBase::isExceptionRaised() const
...
 void BendersBase::UpdateOverallCosts()
 {
-    auto obj = MasterObjectiveFunctionCoeffs();
-    _data.invest_cost = 0;
-    for (const auto& [var_name, var_id]: MasterVariables())
-    {
-        _data.invest_cost += obj[var_id] * _data.x_cut.at(var_name);
-    }
-
+    // The master cost is calculated as the sum of the investment cost and the overall subproblem cost under approximation. Calculating the investment cost this way is more robust than using x_cut, as it ensures that all master variables are accounted for.
```
tbittar:
```suggestion
    // The master cost is calculated as the sum of the investment cost and the overall subproblem cost under approximation. Calculating the investment cost this way is more robust than computing c*x_cut, as it ensures that all master variables are accounted for. Indeed x_cut only contain variables that couple the master and the subproblems and not master only variables (which may have an associated cost).
```

tbittar (changes requested):

src/cpp/benders/benders_by_batch/BendersByBatch.cpp
```diff
@@ -182,15 +182,24 @@ void BendersByBatch::ComputeXCut()
     {
         _data.x_in = _data.x_out;
         _data.x_cut = _data.x_out;
+        _data.non_subpb_vars_in = _data.non_subpb_vars_out;
```
tbittar: I find it clearer not to refer to subpb for data that refers to master pb
```suggestion
        _data.master_only_vars_in = _data.master_only_vars_out;
```

src/cpp/benders/benders_by_batch/BendersByBatch.cpp
```diff
@@ -182,15 +182,24 @@ void BendersByBatch::ComputeXCut()
...
+        _data.non_subpb_vars_cut = _data.non_subpb_vars_out;
```
tbittar:
```suggestion
        _data.master_only_vars_cut = _data.master_only_vars_out;
```

src/cpp/benders/benders_core/include/antares-xpansion/benders/benders_core/WorkerMaster.h
```diff
@@ -82,6 +84,7 @@ class WorkerMaster: public Worker
     void _set_upper_bounds() const;
     void _set_alpha_var();
     void _set_nb_units_var_ids();
+    void _set_non_subproblems_var_ids();
```
tbittar:
```suggestion
    void _set_master_only_var_ids();
```

src/cpp/benders/benders_core/WorkerMaster.cpp
```diff
@@ -95,7 +96,8 @@ void WorkerMaster::restoreFeasibility(std::vector<double>& solution)
  */
 void WorkerMaster::get(Point& x_out,
                        double& overall_subpb_cost_under_approx,
-                       DblVector& single_subpb_costs_under_approx)
+                       DblVector& single_subpb_costs_under_approx,
+                       DblVector& non_subpb_vars_out)
```
tbittar:
```suggestion
                       DblVector& master_only_vars_out)
```

src/cpp/benders/benders_core/WorkerMaster.cpp
```diff
@@ -471,6 +477,34 @@ void WorkerMaster::_set_nb_units_var_ids()
...
+            {
+                unlisted_id = false;
+            }
+        }
+        if (i != _id_alpha && unlisted_id)
+        {
+            _id_non_subpb_vars.push_back(i);
+        }
+    }
+}
```
tbittar: Your approach could be optimized : 
- For each col_id, you iterate over all the two maps, giving a O(2n^2) complexity
- I think we can exploit the fact that variables are added to the master before the alphas: 
    -  Narrow your search from 0 to (n_cols - subproblems_count). (Check if there is a need for -+1 given the alpha var and indexing starting from 0) Add a comment to clearly expose this assumption.
    - Then you can just create the vector `_id_master_only_var` containing initially all int from 0 to n_cols - subproblems_count. Then you iterate over `id_to_name` and erase the `id_to_name[i]` element of `id_master_only`. This way you get your vector with only one vector iteration, complexity O(number_of_candidates). Does that make sense ?

tests/cpp/benders/WorkerMasterTest.cpp
tbittar, line 1: I would indeed create an additionnal unit test (in memory if possible) where you give a VariableMap, subproblems_count, a master problem (with master only vars ie. that are not in the variable map), and check that the id maps are correctly built when calling the master worker constructor. This test may be quite annoying to set up because WorkerMaster is a complicated data structure that needs a lot of thing to work, so you may need to use mocks. I suggest you try to use AI to do this, especially because there are already some tests for it to mimic. If this does not work quickly  do not bother to much

tbittar (changes requested):

tbittar (approved):

PR #1170, opened 2026-01-15
feat: Benders Micro Iterations

hedibouchehda1997:
The number of considered timestamps can get very large which leads to a significant of subproblems. When it comes to modelling the subproblem, we actually consider the grid with N lines and the different N-1 lines cases. Which can lead to heavy optimization problems that are quite time consuming to solve. 

So the idea behind this development proposes a solution to accelerate the solving of subproblems

The idea is to initially solve the subproblem by only considering constraints related to the N lines case. Once the problem is solved, we take its solution and check if it verifies the constraints that haven't been considered while solving. 

One the checks are done, we take the list of violated constraints and we add them to the subproblem then we solve it again. 
Capture d’écran du 2026-01-19 15-58-18

We stop solving the subproblem, once we have no violated conditions. 

Some details about the behaviour : 

When we get from one benders iteration to another, we consider the problem before without the
added constraints. This specification breaks the hot start that is already implemented (the
problem at the end of benders iteration i will not have the same size as the one at the beginning
of benders iteration i+1).
Another implementation consideration is the following: we need to check if the violated
conditions have been added or not before adding it a second time.
Here’s a schema representing the full workflow :

Capture d’écran du 2026-01-19 16-04-32

What will change in an investment study 

**constraints_dictionnary.csv**
This file will allow the mapping between the (line,incident) pair and the constraints associated to it. It has the following structure : 

Line_incident_identifier_1 : constraint_1 , constraint_2 ……
Line_incident_identifier_2 : constraint_1 , constraint_2 ……
.
.
.

**What's used for ?** 

So at every micro-iteration, the julia block will send, if there’s any, a list of violated constraints.
This list will contain the line_incident identifiers. Through these identifiers, we can fetch the
constraints in the associated mps files and add the new rows to the subproblem.
We also have a new repo “constraints” that will contain the mps of constraints associated to
each subproblem. After finding the row names, we fetch for the columns coefficient to add.

**investment_dictionary.csv**

This file will allow the mapping between the line and it corresponding variable name in the optimization problems.

**variables_dictionary.csv**
This file will provide the mapping of each line and the flow that we need to have to be able to verify the constraints at each micro iteration.

**sub repo**

This repo will contains the mps containing the different constraints that hasn't been initially added to subproblem. 

**What changes in the code**

In this PR, we provide the following objects 

**ConstraintsReader**
This class is a wrapper around solver abstract allow reading the constraints mps file and provide the necessary method for the overall Micro iteration workflow.

**BendersPlugin**
The interface to implement when we want to add some plugin behavior

**BendersPluginFactory** 
The object called in BendersFactory to create the BendersPlugin


src/cpp/benders/benders_core/BendersBase.cpp
```diff
@@ -588,15 +593,56 @@ void BendersBase::SolveSubproblem(PlainData::SubProblemData& subproblem_data,
...
+        bool added_rows = true ; 
+        while (added_rows)
+        {
+            std::cout<<"restart solving ....."<<std::endl ; 
+            worker->solve(subproblem_data.lpstatus,
+                _options.OUTPUTROOT,
+                _options.LAST_MASTER_MPS + MPS_SUFFIX,
+                _writer);
+                
+                if (benders_plugin_)
```
JasonMarechal25: issue: If there is no plugin, while loop is infinite

src/cpp/benders/benders_core/BendersBase.cpp
```diff
@@ -588,15 +593,56 @@ void BendersBase::SolveSubproblem(PlainData::SubProblemData& subproblem_data,
...
     Timer subproblem_timer;
     worker->fix_to(_data.x_cut);
-    worker->solve(subproblem_data.lpstatus,
-                  _options.OUTPUTROOT,
-                  _options.LAST_MASTER_MPS + MPS_SUFFIX,
-                  _writer);
+    if (benders_plugin_)
+        benders_plugin_->OnBendersMicroIterationStart() ; 
+
+    if (_options.MICRO_ITERATIONS)
```
JasonMarechal25: suggestion: move into named function

src/cpp/benders/benders_core/BendersBase.cpp
```diff
@@ -588,15 +593,56 @@ void BendersBase::SolveSubproblem(PlainData::SubProblemData& subproblem_data,
...
-                  _options.OUTPUTROOT,
-                  _options.LAST_MASTER_MPS + MPS_SUFFIX,
-                  _writer);
+    if (benders_plugin_)
+        benders_plugin_->OnBendersMicroIterationStart() ; 
+
+    if (_options.MICRO_ITERATIONS) 
+    {    
+        bool added_rows = true ; 
+        while (added_rows)
```
JasonMarechal25, line 638: issue: infinite loop if there is no benders_plugin_

src/cpp/benders/benders_core/ConstraintsReader.cpp
```diff
@@ -0,0 +1,154 @@
...
+    logger_(logger)
+
+{
+
+    SolverFactory solver_factory(logger_) ;
+    subproblem_worker_ = subproblem_worker ; 
+    solver_ = solver_factory.create_solver(solver_name,SOLVER_TYPE::CONTINUOUS,solver_log_manager) ; 
+    if (solver_)
+    {
+        solver_->set_threads(1) ;
```
JasonMarechal25: question: Is it necessary. If yes please document why, maybe just a comment.

src/cpp/benders/benders_core/ConstraintsReader.cpp
```diff
@@ -0,0 +1,154 @@
...
+    }
+    return row_pos ; 
+}
+
+constraintRow ConstraintsReader::get_row(const std::string& name) 
+{
+    constraintRow result ; 
+    result.range_p = {} ; 
+    result.row_names = {name} ; 
+    int constraint_pos = get_row_index(name) ;
```
JasonMarechal25: nitpick: move inside _if_

src/cpp/benders/benders_core/ConstraintsReader.cpp
```diff
@@ -0,0 +1,154 @@
...
+                    const std::filesystem::path variables_names_path,
+                    const std::shared_ptr<SubproblemWorker>& subproblem_worker) : 
+    logger_(logger)
+
+{
+
+    SolverFactory solver_factory(logger_) ;
+    subproblem_worker_ = subproblem_worker ; 
+    solver_ = solver_factory.create_solver(solver_name,SOLVER_TYPE::CONTINUOUS,solver_log_manager) ; 
+    if (solver_)
```
JasonMarechal25: question: does it make any sense to have a null solver?

tbittar:
The PR description is great to understand the overall idea of the PR. I think that we should start from it to write proper documentation, see https://github.com/AntaresSimulatorTeam/antares-xpansion/issues/1178, trying to clearly separate the description of the generic way of working with plugins and the practical example of this PR

src/cpp/benders/benders_core/include/antares-xpansion/benders/benders_core/ConstraintsReader.h
```diff
@@ -0,0 +1,59 @@
+#pragma once 
+
+#include "IBendersProblemProvider.h"
+#include "antares-xpansion/benders/benders_core/BendersProblemFromFile.h"
+#include "antares-xpansion/benders/benders_core/SubproblemWorker.h"
+#include <antares-xpansion/benders/benders_core/SolverIO.h>
+#include <utility>
+
+
+struct constraintRow
```
tbittar: My name suggestion is not really good, but the idea is that you have a struct that stores rows as they are in the underlying solver
```suggestion
struct SolverRepresentedRows
```

src/cpp/benders/benders_core/include/antares-xpansion/benders/benders_core/SubproblemWorker.h
```diff
@@ -27,9 +27,11 @@ class SubproblemWorker: public Worker
                      double cut_coefficient_tolerance);
     virtual ~SubproblemWorker() = default;
     std::vector<double> get_solution() const;
+    int get_variable_index(const std::string& variable_name) ;
```
tbittar: This function is a generic solver function, I do not think it belongs to `SubproblemWorker`. Is it not sufficient to do something like `subproblem_worker->solver->get_var_index()` ?

src/cpp/benders/benders_core/BendersBase.cpp
```diff
@@ -605,15 +607,44 @@ void BendersBase::SolveSubproblem(PlainData::SubProblemData& subproblem_data,
 {
     Timer subproblem_timer;
     worker->fix_to(_data.x_cut);
-    worker->solve(subproblem_data.lpstatus,
-                  _options.OUTPUTROOT,
-                  _options.LAST_MASTER_MPS + MPS_SUFFIX,
-                  _writer);
+    if (benders_plugin_)
```
tbittar: Given the name of the call back , I think we do not want to call it f the option is not activated
```suggestion
    if (benders_plugin_ && _options.MICRO_ITERATIONS)
```

src/cpp/benders/benders_core/ConstraintsReader.cpp
tbittar: I do not think so, and an error at solver creation would be catched within `create_solver`

src/cpp/benders/benders_core/ConstraintsReader.cpp
```diff
@@ -0,0 +1,154 @@
...
+    {
+        solver_->set_threads(1) ; 
+        solver_->set_output_log_level(log_level);
+        benders_problem_provider_ = std::make_shared<BendersProblemFromFile>(constraint_file_path) ; 
+        solver_IO_.configure(solver_name,ProblemsFormat::MPS_FILE) ;
```
tbittar: The interface is hard-coded to be the MPS format. Would it be possible to at least accept both MPS / SVF ? Is even a sensible question given how these files will be generated in Julia (needs same Xpress version, ...)

src/cpp/benders/benders_core/ConstraintsReader.cpp
```diff
@@ -0,0 +1,154 @@
...
+    solver_ = solver_factory.create_solver(solver_name,SOLVER_TYPE::CONTINUOUS,solver_log_manager) ; 
+    if (solver_)
+    {
+        solver_->set_threads(1) ; 
+        solver_->set_output_log_level(log_level);
+        benders_problem_provider_ = std::make_shared<BendersProblemFromFile>(constraint_file_path) ; 
+        solver_IO_.configure(solver_name,ProblemsFormat::MPS_FILE) ; 
+        benders_problem_provider_->provide_problem(solver_IO_,solver_) ;
+        int n_rows = solver_->get_nrows() ; 
+        std::cout<<"number of rows in constraint problem "<<n_rows<<std::endl ;
```
tbittar: Clean before merge, or use the logger

src/cpp/benders/benders_core/include/antares-xpansion/benders/benders_core/ConstraintsReader.h
```diff
@@ -0,0 +1,59 @@
...
+    std::vector<int> mclind ; 
+    std::vector<double> dmatval ; 
+    std::vector<double> range_p ; 
+    std::vector<char> qrtype_p ; 
+    std::vector<double> rhs ; 
+    std::vector<std::string> row_names ; 
+} ; 
+
+
+class ConstraintsReader
```
tbittar: I do not completely understand the responsibility of this class: it is named `ConsrtaintReader` but it also adds constraints in the problems. 

Generally I think this class has methods that do not need to be exposed here as they are generic solver methods (such as `get_row_index`, `get_sub_solution` ...) that are directly accessible through the solver object with `subproblem_worker->solver->get_xxx()`

Maybe a separate class for the `add` methods could be cleaner, which would be a wrapper around solver `add` methods using the data you get from the `ConstraintReader`

src/cpp/benders/benders_core/ConstraintsReader.cpp
```diff
@@ -0,0 +1,154 @@
...
+{
+    int row_pos(-1) ; 
+    if (solver_) 
+    {
+        row_pos = solver_->get_row_index(name) ; 
+    }
+    return row_pos ; 
+}
+
+constraintRow ConstraintsReader::get_row(const std::string& name)
```
tbittar: This should be a `SolverAbstract` method. What you do is read a row a store in with the solver representation

src/cpp/benders/benders_core/CouplingMapGenerator.cpp
```diff
@@ -54,3 +54,24 @@ CouplingMap CouplingMapGenerator::BuildInput(const std::filesystem::path& struct
...
+{
+    for (auto&& [subProblemName,variable_map] : coupling_map) 
+    {
+        size_t underscore_pos = subProblemName.find('_') ; 
+        size_t dot_pos =  subProblemName.find('.') ; 
+
+        std::string subproblem_num = subProblemName.substr(underscore_pos + 1, dot_pos - underscore_pos - 1);
+        if (subproblem_num != "master") 
+        {
+            std::string constraint_str = "constraints/constraints_" + subproblem_num + ".mps" ;
```
tbittar: Here we heavily rely on a given way to generate these constraints problems... We need something more robust (input parameter ?)

src/cpp/benders/benders_core/SubproblemWorker.cpp
```diff
@@ -109,3 +109,13 @@ std::vector<double> SubproblemWorker::get_solution() const
     }
     return solution;
 }
+
+int SubproblemWorker::get_variable_index(const std::string& variable_name)
```
tbittar: not needed

src/cpp/benders/benders_mpi/BendersMPI.cpp
```diff
@@ -404,16 +407,22 @@ void BendersMpi::Run()
...
-
+        
         /*Solve Master problem, get optimal value and cost and send it to
-         * process*/
-        step_1_solve_master();
+        * process*/
+       step_1_solve_master();
+
+       if (benders_plugin_)
+            benders_plugin_->OnBendersMasterIterationStart(_data.x_out) ;
```
tbittar:
```suggestion
            benders_plugin_->OnMasterResolutionEnd(_data.x_out) ; 
```

src/cpp/benders/benders_mpi/BendersMPI.cpp
```diff
@@ -430,6 +439,10 @@ void BendersMpi::Run()
             mathLoggerDriver_->Print(_data);
             SaveCurrentBendersData();
         }
+    
+        if (benders_plugin_)
+            benders_plugin_->OnBendersMasterIterationEnd() ;
```
tbittar:
```suggestion
            benders_plugin_->OnBendersIterationEnd() ; 
```

src/cpp/benders/factories/BendersPluginFactory.cpp
```diff
@@ -0,0 +1,46 @@
...
+
+BendersPluginFactory::BendersPluginFactory(const std::filesystem::path& input_root) 
+
+{
+    library_path_ = input_root / "libmylib/lib/libmylib.so" ; 
+    input_root_ = input_root ; 
+}
+
+
+BendersPlugin* BendersPluginFactory::CreatePlugin(const CouplingMap& coupling_map,bool micro_iter)
```
tbittar: Later on, we may have plugins (which is a generic concept) that are not linked to the use case of micro iterations. So I do not see the point of having the `micor_iter` argument here. I would rather have a higher level variable (in Bendersfactory) that says whether or not we need a benders plugin factory and create some. And within this function getting rid of the if condition

src/cpp/benders/plugins/include/antares-xpansion/benders/plugins/Benders_Jl_MICRO_ITERS.h
```diff
@@ -0,0 +1,103 @@
...
+
+
+struct SubProblemIds 
+{
+    const char** subProblems_ids ; 
+    int n_subproblems ; 
+} ; 
+
+
+struct CandidateLineMasterIterationResult
```
tbittar: It seems that in this file, we are on the "business domain" rather than "mathematical domain", as all of this is specific to the investment on detailed network. So we need to use business names
```suggestion
struct CandidateLineInvestmentStatus
```

src/cpp/benders/plugins/include/antares-xpansion/benders/plugins/Benders_Jl_MICRO_ITERS.h
```diff
@@ -0,0 +1,103 @@
...
+
+
+struct CandidateLineMasterIterationResult 
+{
+    const char* candidate_line_id ; 
+    int is_invested ; 
+}; 
+
+
+struct MasterBendersInput
```
tbittar:
```suggestion
struct CandidateLineInvestmentStatusList
```

src/cpp/benders/plugins/include/antares-xpansion/benders/plugins/Benders_Jl_MICRO_ITERS.h
```diff
@@ -0,0 +1,103 @@
...
+} ; 
+
+struct FlowNList 
+{
+    FlowN* flows ; 
+    int size ; 
+} ; 
+
+
+struct ConstraintsToAdd
```
tbittar:
```suggestion
struct ViolatedFlowConstraints
```

src/cpp/benders/plugins/Benders_Jl_MICRO_ITERS.cpp
```diff
@@ -0,0 +1,290 @@
...
+    coupling_map_ = coupling_map ; 
+    CouplingMapGenerator::BuildSubProblemConstaintMap(coupling_map_,subproblem_constraint_map_,constraints_coupling_map_) ; 
+
+
+    input_root_ = input_root ; 
+
+    std::filesystem::path mirco_iterations_options_path = input_root_ / "micro_iterations_config.txt" ; 
+    std::ifstream micro_iterations_options_stream(mirco_iterations_options_path.c_str()) ; 
+
+    if (micro_iterations_options_stream.is_open())
```
tbittar: Introduce a function for each file you read to improve the readbility of this function

tbittar (changes requested):

src/cpp/benders/benders_core/BendersBase.cpp
hedibouchehda1997: if Micro Iterations is true and there's no plugin the code exit fail before launching benders

src/cpp/benders/benders_core/ConstraintsReader.cpp
hedibouchehda1997: the set_threads fixes the number of threads on which we're solving the optimization problems. Maybe not setting it will make the solver run on multiple threads which can problematic for the cpu task manager since we already working with  mpi (avoid any multithreading in that context)

src/cpp/benders/benders_core/include/antares-xpansion/benders/benders_core/SubproblemWorker.h
hedibouchehda1997: one OOP good practice is to never call class members out of class method. So this function is just there to respect the encapsulation principle

src/cpp/benders/benders_core/include/antares-xpansion/benders/benders_core/ConstraintsReader.h
```diff
@@ -1,59 +1,77 @@
...
 #include "IBendersProblemProvider.h"
 #include "antares-xpansion/benders/benders_core/BendersProblemFromFile.h"
 #include "antares-xpansion/benders/benders_core/SubproblemWorker.h"
-#include <antares-xpansion/benders/benders_core/SolverIO.h>
-#include <utility>
 
 
-struct constraintRow 
+/*
+    The structure contains the necesary vector to fetch for a row of an mps file
```
tbittar: necessary

src/cpp/benders/benders_core/include/antares-xpansion/benders/benders_core/ConstraintsReader.h
```diff
@@ -1,59 +1,77 @@
...
+public:
+
+    /*
+        Constructor 
+        @inputs : 
+            - constraint_file_path : path to the mps constraint file
+            - solver_name : solver name 
+            - solver_log_manager : solver log manager 
+            - logger : benders logger 
+            - log_levl : logging level
```
tbittar: log_level

src/cpp/benders/benders_mpi/BendersMPI.cpp
```diff
@@ -403,26 +401,30 @@ void BendersMpi::Run()
...
+
         /*Solve Master problem, get optimal value and cost and send it to
-        * process*/
-       step_1_solve_master();
+         * process*/
+        step_1_solve_master();
 
-       if (benders_plugin_)
-            benders_plugin_->OnBendersMasterIterationStart(_data.x_out) ; 
+        broadcast(_world,_data.x_out,rank_0) ;
```
tbittar: There is already a broadcast of `x_cut` here: https://github.com/AntaresSimulatorTeam/antares-xpansion/blob/b742824fd6d31f1c0a7cb409f160df22a38cdc95/src/cpp/benders/benders_mpi/BendersMPI.cpp#L121-L122 , and you should solve the subproblems / do the micro iterations at `x_cut`, not `x_out`

src/cpp/benders/benders_mpi/BendersMPI.cpp
```diff
@@ -403,26 +401,30 @@ void BendersMpi::Run()
...
+         * process*/
+        step_1_solve_master();
 
-       if (benders_plugin_)
-            benders_plugin_->OnBendersMasterIterationStart(_data.x_out) ; 
+        broadcast(_world,_data.x_out,rank_0) ; 
 
+        if (benders_plugin_ )
+        {
+            benders_plugin_->OnBendersMasterIterationStart(_data.x_out,_data.it);
```
tbittar: _data.x_cut

src/cpp/benders/logger/include/antares-xpansion/benders/logger/MicroIterationsLog.h
```diff
@@ -0,0 +1,114 @@
+//This header file will contain all the necessary objects to handle micro iterations workflow loggin
```
tbittar: logging

src/cpp/benders/logger/include/antares-xpansion/benders/logger/MicroIterationsLog.h
```diff
@@ -0,0 +1,114 @@
...
+#include "antares-xpansion/benders/benders_core/SimulationOptions.h"
+
+
+/*
+This structure contains the master iteration data that we want want to log 
+    - num_iter : number of master iteration
+    - PTDF_compute_time : elapsed time to compute the new PTDF
+    - removing_rows_per_sub_time : elapsed time to remove the constraints added to subproblem workers 
+*/
+struct MasterIterationLog
```
tbittar: Need to find another name, not to be confused with the master problem resolution

src/cpp/benders/logger/MicroIterationsLog.cpp
```diff
@@ -0,0 +1,123 @@
...
+
+void MicroIterationsLog::Dump()
+{
+        
+        std::filesystem::path micro_iterations_log_path = std::filesystem::path(options_.OUTPUTROOT)  / "micro_iterations.log" ; 
+        std::ofstream micro_iterations_log_stream(micro_iterations_log_path.c_str()) ; 
+        
+        micro_iterations_log_stream<<"************************** MICRO ITERS config ************************** \n\n" ; 
+        if (warm_start_)
+            micro_iterations_log_stream<<"warl_start=1\n\n";
```
tbittar:
```suggestion
            micro_iterations_log_stream<<"warm_start=1\n\n"; 
```

src/cpp/benders/logger/MicroIterationsLog.cpp
```diff
@@ -0,0 +1,123 @@
...
+        
+        for (size_t i=0; i<master_iterations_logs_.size(); i++)
+        {
+            micro_iterations_log_stream<<"Master Iteration : "<<master_iterations_logs_[i].num_iter<<" PTDF compute time : "
+            <<master_iterations_logs_[i].PTDF_compute_time<<"\n" ; 
+            
+            if (!warm_start_)
+            {
+                micro_iterations_log_stream<<"deleting rows time per subproblems\n" ;
+                micro_iterations_log_stream<<"sub       time(us)\n" ;
```
tbittar: In some other log files we have durations either in seconds or milliseconds, I think it is better to stick to milliseconds here

src/cpp/benders/logger/MicroIterationsLog.cpp
```diff
@@ -0,0 +1,123 @@
...
+            micro_iterations_log_stream<<"\n************************** MICRO ITERS INFOS *********************\n\n" ; 
+            micro_iterations_log_stream<<"Micro iter         solving time(us)       n added         adding time(us) \n" ; 
+
+            
+            auto micro_iterations_per_master_iter_per_sub = micro_iterations_per_benders_iter[i] ; 
+            
+            for (auto& [sub_name, micro_iters_vec] : micro_iterations_per_master_iter_per_sub)
+            {  
+                size_t start = sub_name.find_last_of('/') + 1;
+                size_t end = sub_name.find(".mps");
```
tbittar: This heavily relies on the input format which may also be .svf...

src/cpp/benders/benders_core/CouplingMapGenerator.cpp
```diff
@@ -54,3 +54,24 @@ CouplingMap CouplingMapGenerator::BuildInput(const std::filesystem::path& struct
...
+    for (auto&& [subProblemName, variable_map]: coupling_map)
+    {
+        size_t underscore_pos = subProblemName.find('_');
+        size_t dot_pos = subProblemName.find('.');
+
+        std::string subproblem_num = subProblemName.substr(underscore_pos + 1,
+                                                           dot_pos - underscore_pos - 1);
+        if (subproblem_num != "master")
+        {
+            std::string constraint_str = "constraints/constraints_" + subproblem_num + ".mps";
```
tbittar: Heavily based on mps format and even on the way the constraints_coupling_map has been generated

src/cpp/benders/plugins/Benders_Jl_MICRO_ITERS.cpp
```diff
@@ -0,0 +1,389 @@
...
+    CouplingMapGenerator::BuildSubProblemConstaintMap(coupling_map_,
+                                                      subproblem_constraint_map_,
+                                                      constraints_coupling_map_);
+
+    input_root_ = options_.INPUTROOT ;
+    warm_start_ = true ; 
+
+    //Reading the micro iterations configuration file
+    std::filesystem::path mirco_iterations_options_path = input_root_
+                                                          / "micro_iterations_config.txt";
```
tbittar, line 147: Why not having a more structured input file format like yaml or json ? This would make file creatoin easier for user and parsing also

src/cpp/benders/plugins/Benders_Jl_MICRO_ITERS.cpp
```diff
@@ -0,0 +1,389 @@
...
+    std::ifstream micro_iterations_options_stream(mirco_iterations_options_path.c_str());
+    
+
+    if (micro_iterations_options_stream.is_open())
```
tbittar, line 150: Factorize each file you read in a separate function to improve readability

src/cpp/benders/plugins/Benders_Jl_MICRO_ITERS.cpp
```diff
@@ -0,0 +1,389 @@
...
+    {
+        std::cerr<<"unable to open : "<<libmylib_path.c_str()<<std::endl; 
+        exit(EXIT_FAILURE) ; 
+    }
+
+    if (options_.LOG_LEVEL >= 2)
+        micro_iterations_logger_ = std::make_shared<MicroIterationsLog>(options_,subproblem_constraint_map_,constraints_csv_map_,warm_start_) ; 
+}
+
+Benders_Jl_MICRO_ITERS::~Benders_Jl_MICRO_ITERS()
```
tbittar: You cannot just use the default destructor ?

src/cpp/benders/plugins/Benders_Jl_MICRO_ITERS.cpp
```diff
@@ -0,0 +1,389 @@
...
+                                            const BendersBaseOptions& options,
+                                            const SolverLogManager& solver_log_manager)
+{
+    _logger = logger;
+
+    
+    
+    if (handle_)
+    {
+        int size_subs = sub_pb_ids_.n_subproblems;
```
tbittar: Where is this used ?

src/cpp/benders/plugins/include/antares-xpansion/benders/plugins/Benders_Jl_MICRO_ITERS.h
```diff
@@ -0,0 +1,266 @@
...
+
+
+/*
+    This structure will be used after solving the master problem. 
+    @members : 
+        - candidate_line_id : id of the candidate line 
+        - is_invested : 0 if not invest, 1 if invested
+
+*/
+struct CandidateLineMasterIterationResult
```
tbittar:
```suggestion
struct CandidateLineInvestmentStatus
```

src/cpp/benders/benders_core/SubproblemWorker.cpp
```diff
@@ -109,3 +109,31 @@ std::vector<double> SubproblemWorker::get_solution() const
     }
     return solution;
 }
+
+
+void SubproblemWorker::delete_rows(int start_pos)
```
tbittar:
```suggestion
void SubproblemWorker::delete_last_rows(int start_pos) 
```

tbittar (changes requested):

src/cpp/benders/plugins/Benders_Jl_MICRO_ITERS.cpp
hedibouchehda1997: need to set subs in julia code

src/cpp/benders/benders_core/include/antares-xpansion/benders/benders_core/BendersBase.h
```diff
@@ -56,6 +58,9 @@ class BendersBase
     void set_input_map(const CouplingMap& coupling_map);
     int MasterRowIndex(const std::string& row_name) const;
     void MasterChangeRhs(int id_row, double val) const;
+    // void set_subproblem_constraint_map(const SubProblemConstraintMap& subproblem_constraint_map,
```
JasonMarechal25: issue: remove commented code

src/cpp/benders/plugins/include/antares-xpansion/benders/plugins/BendersPlugin.h
```diff
@@ -0,0 +1,47 @@
+#pragma once
+
+#include <map>
```
JasonMarechal25: issue: remove superfluous includes

JasonMarechal25 (changes requested):

src/cpp/benders/factories/BendersFactory.cpp
```diff
@@ -125,8 +125,8 @@ auto BendersFactory::ConfigureBenders(const BendersBaseOptions& benders_options,
     }
 
     std::shared_ptr<BendersPlugin> benders_plugin(
-      benders_plugin_factory_->CreatePlugin(coupling_map, options_.MICRO_ITERATIONS));
-
+      benders_plugin_factory_->CreatePlugin(coupling_map, options_.MICRO_ITERATIONS,world_));
+    std::cout<<"benders pluging factory !!!"<<std::endl ;
```
tbittar: Remove if debug print

tbittar (approved):

JasonMarechal25 (approved):

PR #1176, opened 2026-01-20
Fixing subs per cut 

hedibouchehda1997:
Before this dev , the subproblem considered in each cut can be different with the number of process we consider when we run benders which lead to different number of iterations of the master problem (dependending on the chosen nproc).
This dev will make the cut robust and make sure that we always consider the same subproblems at each cut

### DIFF
--- a/src/cpp/benders/benders_core/BendersBase.cpp
+++ b/src/cpp/benders/benders_core/BendersBase.cpp
@@ -734,34 +734,51 @@ std::vector<SubProblemNamesInCut> BendersBase::split_subproblem_data_pairs(
   int max_aggregation) const
 {
     int n_cuts = SetAggregation(max_aggregation);
-    std::vector<SubProblemNamesInCut> result(n_cuts);
-
-    if (_data.nsubproblem == 0 || n_cuts <= 0)
+    struct Entry
     {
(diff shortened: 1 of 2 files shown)


tbittar (commented):
Is it possible / not too complicated to write a unit test ?

src/cpp/benders/benders_core/BendersBase.cpp
```diff
@@ -734,34 +734,56 @@ std::vector<SubProblemNamesInCut> BendersBase::split_subproblem_data_pairs(
...
-    if (_data.nsubproblem == 0 || n_cuts <= 0)
+    struct Entry
     {
-        return std::vector<std::vector<std::pair<std::string, int>>>();
-    }
-
-    size_t target_per_cut = (_data.nsubproblem + n_cuts - 1) / n_cuts;
+        const std::string* name = nullptr;
+        int vecPos = -1;
+    };
```
tbittar: This is defined within a function, does that mean this is redeclared (and some operations are done) each time this function is called @JasonMarechal25 ?

hedibouchehda1997: better to get it out of the function. i'll do that !

PR #1192, opened 2026-02-02
feat(problem-generation): enhance simulation handling

JasonMarechal25:
* Use Antares API to test if weeks are independent
- yes : Call SingleProblemGetter API point to avoid doing simulation
- no : Call API (v1) to perform simulation


tbittar (approved):

antares-version.json
```diff
@@ -1,5 +1,5 @@
 {
-  "antares_version": "9.3.4",
+  "antares_version": "9.3.6-rc4",
   "antares_version_executable": "9.3",
```
flomnes:
```suggestion
  "antares_version": "9.3.6",
"antares_version_executable": "9.3",
```

src/cpp/lpnamer/main/ProblemGeneration.cpp
```diff
@@ -3,10 +3,12 @@
 #include <antares-xpansion/lpnamer/problem_modifier/StructureGeneration.h>
 #include <execution>
 #include <iostream>
-#include <tbb/tbb.h>
+#include <tbb/tbb.h> //Do not remove
 #include <utility>
 
 #include <antares/api/singleProblemGetter.h>
+#include <antares/api/solver.h>
+#include "antares/file-tree-study-loader/FileTreeStudyLoader.h"
```
flomnes:
```suggestion
```

src/cpp/lpnamer/main/ProblemGeneration.cpp
```diff
@@ -86,6 +159,26 @@ void ProblemGeneration::performAntaresSimulation(const std::filesystem::path& st
...
+  const std::filesystem::path& study_dir,
+  const std::filesystem::path& simulation_dir,
+  ProblemGenerationLog::ProblemGenerationLogger* logger)
+{
+    Antares::Solver::SingleProblemGetter spg(study_dir);
+    if (spg.areWeeksIndependent())
+    {
+        (*logger)(LogUtils::LOGLEVEL::INFO)
+          << "Weeks are independent, using optimized problem generation" << std::endl;
+        generate_antares_problems(study_dir, simulation_dir);
```
flomnes: In this branch the study is loaded twice, which is a waste of time and I/O. To avoid this, I suggest passing `spg` to `generate_antares_problems`.

src/cpp/lpnamer/main/ProblemGeneration.cpp
```diff
@@ -86,6 +159,26 @@ void ProblemGeneration::performAntaresSimulation(const std::filesystem::path& st
...
+    }
+    else
+    {
+        (*logger)(LogUtils::LOGLEVEL::INFO)
+          << "Weeks are dependent, performing full Antares simulation" << std::endl;
+        performAntaresSimulation(simulation_dir);
```
flomnes, line 174: Same problem in this branch, the study is loaded twice. No obvious solution though

flomnes (dismissed):

src/cpp/lpnamer/main/ProblemGeneration.cpp
tbittar: From an operational perspective, this situation is lesslikely than the other one and we will pay the simulation cost. I think we can afford the cost of loading the simulation twice (much smaller than simulation solving time anyway....). Just be sure than the memory is correctly cleared after first loading...

src/cpp/lpnamer/main/ProblemGeneration.cpp
JasonMarechal25: Nice catch. Fixed

src/cpp/lpnamer/main/ProblemGeneration.cpp
JasonMarechal25: For now I opted to leave it as it is. Loading the study twice is far from optimal but probably "negligible" compared to simulation time.

With the solution of a "meta" API in simulator it will probably solve itself

PR #1198, opened 2026-02-10
workflow Antares-Problem-Generator -> Benders

a-zakir:


tbittar:
Nice to quickly test the workflow, a few comments:
- I understand that the user should directly generate the `options.json` and pass it via command line. That is we do not use `settings.ini` anymore. Although this seems natural and simpler from benders perspective, need to evaluate if this is ok (especially for integration with AntaresWeb and user workflow to generate studies) -> In the short term we may need to rely on `settings.ini` for quicker integration...
- Also this workflow does not handle weighted years (same remark applies: should the user pass directly the file with correct "benders" format to command line or we rely on the current format (which is processed within problem generation... a bit annoying))

PR #1202, opened 2026-02-12
Feature/revert cut agregation try 2

tbittar:


tbittar:
Cut aggregation is now fully functional

PR #1205, opened 2026-02-13
Benders Plugin

hedibouchehda1997:
Here we introduce the Benders Plugin mechanism that will be used in Micro iteration worlflow, but can be generalized.

The idea is to provide callback functions at different points of the benders workflow (at start, at end, at the beginning and the end of the master iteration and at the beginning and the end of micro iteration for each subproblem)


src/cpp/benders/benders_core/BendersBase.cpp
```diff
@@ -612,6 +612,12 @@ void BendersBase::SolveSubproblem(PlainData::SubProblemData& subproblem_data,
 {
     Timer subproblem_timer;
     worker->fix_to(_data.x_cut);
+
+    if (benders_plugin_)
+    {
+        benders_plugin_->OnBendersMicroIterationStart();
+    }
```
a-zakir: i think you can totally avoid ```if (benders_plugin_)``` check by providing default no-op plugin instead....

src/cpp/benders/benders_mpi/BendersMPI.cpp
```diff
@@ -471,9 +481,20 @@ void BendersMpi::launch()
...
+    {
+        benders_plugin_->OnBendersStart(subproblem_map, _logger, _options, solver_log_manager_);
+    }
+
     Run();
     _world.barrier();
 
+    if (benders_plugin_)
+    {
+        int rank_proc = Rank();
```
a-zakir: since there is one plugin instance per proc, the proc number can be passed via the plugin constructor

src/cpp/benders/benders_core/BendersBase.cpp
```diff
@@ -620,6 +626,12 @@ void BendersBase::SolveSubproblem(PlainData::SubProblemData& subproblem_data,
     worker->get_subgradient(subproblem_data.var_name_and_subgradient);
     worker->get_splex_num_of_ite_last(subproblem_data.simplex_iter);
 
+    if (benders_plugin_)
+    {
+        benders_plugin_->OnBendersMicroIterationStart();
```
a-zakir: why OnBendersMicroIterationStart is called twice  before/after solve?

src/cpp/benders/benders_core/BendersBase.cpp
tbittar: This means that benders will always need to use a plugin (a no-op one in the default case, that is in most of our use cases), is this desirable ? And we need to provide its implementation within this repo ?

src/cpp/benders/benders_mpi/BendersMPI.cpp
```diff
@@ -409,6 +409,11 @@ void BendersMpi::Run()
          * process*/
         step_1_solve_master();
 
+        if (benders_plugin_)
+        {
+            benders_plugin_->OnBendersMasterIterationStart(_data.x_cut, _data.it);
```
tbittar: We call this after having solved the master
```suggestion
            benders_plugin_->OnBendersMasterResolutionEnd(_data.x_cut, _data.it);
```

src/cpp/benders/benders_mpi/BendersMPI.cpp
```diff
@@ -430,6 +435,11 @@ void BendersMpi::Run()
             mathLoggerDriver_->Print(_data);
             SaveCurrentBendersData();
         }
+
+        if (benders_plugin_)
+        {
+            benders_plugin_->OnBendersMasterIterationEnd();
```
tbittar: Here we are the end of a "regular" Benders iteration
```suggestion
            benders_plugin_->OnBendersIterationEnd();
```

tbittar (changes requested):

src/cpp/benders/benders_core/BendersBase.cpp
JasonMarechal25: It's a mean to avoid "if plugin" everywhere.
Plugin factory will try to load a plugin. If none exists it will return a "NoOperationPlugin" for which all method will fallthrough.

a-zakir: this kind of behavior exist in Benders, e.g logs, for (**rank >0**) logs do nothing

src/cpp/benders/plugins/include/antares-xpansion/benders/plugins/BendersPlugin.h
```diff
@@ -0,0 +1,71 @@
...
+    virtual void OnBendersMicroIterationStart() = 0;
+
+    /*
+      This method will be called after solving a subproblem (for each subproblem)
+      @inputs :
+            - sub_name : subproblem name
+            - added_rows : if any rows we have to add to the subproblem worker
+            - solve_time : elapsed time to solve the subproblem
+    */
+    virtual void OnBendersMicroIterationEnd(std::string sub_name,
```
JasonMarechal25: thought; The interface could be clearer.
Reading this I understand:
I am Benders, I send the message OnBendersMicroIterationEnd for subproblem _sub_name_, solve took _solve_time_. Plugin will do something that potentialy add rows and set _added_rows_ flag.

It is not clear why solve_time is necessary

Added_row flag exists but not add_row column. It seems adherent to your Julia plugin implementation. As much as possible, benders shouldn't need to react to what happens after OnCallbacks. If the behaviour of benders need to be altered, plugin implementation should call method on benders and set flag, value, or data appropriately.

JasonMarechal25 (approved):

src/cpp/benders/benders_mpi/BendersMPI.cpp
a-zakir: We don’t need to store this info.

src/cpp/benders/benders_mpi/BendersMPI.cpp
hedibouchehda1997: actually the inputs of all the plugin methods are designed for the micro iterations case which is not general enough for other use cases

src/cpp/benders/plugins/include/antares-xpansion/benders/plugins/BendersPlugin.h
hedibouchehda1997: since the inputs are too specific to the micro iterations case. For this PR, i think it's better to merge with methods that have no input. then we will adapt them during the micro iterations PR

tbittar (approved):

JasonMarechal25 (approved):

JasonMarechal25:
LTGM
@hedibouchehda1997 just need to fix segfault visible in tests

src/cpp/benders/benders_mpi/BendersMPI.cpp
```diff
@@ -430,6 +436,8 @@ void BendersMpi::Run()
             mathLoggerDriver_->Print(_data);
             SaveCurrentBendersData();
         }
+
+        benders_plugin_->OnBendersMasterIterationEnd();
```
tbittar: Why 2 calls of `benders_plugin_->OnBendersMasterIterationEnd();` ?

tbittar: Also please take into account the renaming suggestion : https://github.com/AntaresSimulatorTeam/antares-xpansion/pull/1205#discussion_r2815862007

src/cpp/benders/factories/include/antares-xpansion/benders/factories/BendersPluginFactory.h
```diff
@@ -0,0 +1,43 @@
...
+            - options : study simulation
+    */
+    BendersPluginFactory(const SimulationOptions& options);
+
+    /*
+        This method will be called to instantiate the benders plugin
+        @inputs :
+            - coupling_map : coupling map
+            - micro_iter : boolean to check if Micro iterations is needed to build the right plugin
+            - worlld : MPI communicator
```
tbittar:
```suggestion
            - world : MPI communicator
```

tests/cpp/outer_loop/outer_loop_test.cpp
```diff
@@ -117,19 +119,29 @@ void CheckMinInvestmentConstraint(const VariableMap& master_variables,
...
     CouplingMap coupling_map = CouplingMapGenerator::BuildInput(
-      std::filesystem::path(bendersoptions.INPUTROOT) / bendersoptions.STRUCTURE_FILE,
-      logger.get(),
-      ::testing::UnitTest::GetInstance()->current_test_info()->name());
-    // override solver
+        std::filesystem::path(bendersoptions.INPUTROOT) / bendersoptions.STRUCTURE_FILE,
+        logger.get(),
+        ::testing::UnitTest::GetInstance()->current_test_info()->name());
+        // override solver
+        // std::shared_ptr<BendersPlugin> benders_plugin(benders_plugin_factory_->CreatePlugin(coupling_map, false, pworld))
```
tbittar: remove commented code

src/cpp/benders/benders_mpi/BendersMPI.cpp
ze-hedi: > Why 2 calls of `benders_plugin_->OnBendersMasterIterationEnd();` ?

where the second call is ?

tbittar (approved):

PR #1215, opened 2026-03-13
Fixing alphas in cuts in Aggregation mode

ze-hedi:
This PR intend to enhance the Aggregation mode of Benders in MPI and by batch cases.

So the goal of a Benders simulation is ti minimize the objective function of the master problem : 

image

where theta_i correspond to an estimation of the solution of the ith subproblem.

In Aggregation mode, at each benders iteration, we add constraints, to the master problems, of the following from : 

image

where each theta correspond to an estimation of subproblems we consider in the cut. 

In order to enhance the numerical robustness of the algorithm, we fix the estimation of the solution of the subproblems that we have in one cut : 

image

 And this is what has been implemented in this PR


src/cpp/benders/benders_by_batch/BendersByBatch.cpp
```diff
@@ -5,13 +5,28 @@
 
 #include "antares-xpansion/benders/benders_by_batch/BatchCollection.h"
 #include "antares-xpansion/benders/benders_by_batch/RandomBatchShuffler.h"
+#include "antares-xpansion/benders/benders_core/BendersProblemFromFile.h"
```
tbittar: Where is this used ?

src/cpp/benders/benders_by_batch/BendersByBatch.cpp
```diff
@@ -5,13 +5,28 @@
...
+    BroadCastVariablesIndices();
+    init_problems_ = false;
+}
 
+void BendersByBatch::BuildMasterProblem()
+{
+    ResetMaster();
+    if (_world.rank() == rank_0)
+    {
+        _master->addCutsAlphas(batch_collection_full_for_cuts_);
```
tbittar: Although this function is clearly related to the building of the master problem, it needs information from the batches. Then as is, WorkerMaster becomes dependent of BatchCollection (which is in the benders_by_batch lib) and this is not natural for me.

- Either put this function in BendersByBatch (and its other implementation in BendersMPI)
- Or BatchCollection is in fact only a data container, it can be a utils function outside the Benders by batch library (although it is mainly use for the Benders by batch case...)

I do not really know what is the best, any (other) idea @JasonMarechal25 ?

src/cpp/benders/benders_by_batch/BendersByBatch.cpp
```diff
@@ -5,13 +5,28 @@
...
-
+    BuildBatches();
     BuildMasterProblem();
+    BroadCastVariablesIndices();
+    init_problems_ = false;
+}
 
+void BendersByBatch::BuildMasterProblem()
+{
+    ResetMaster();
```
tbittar: I think `reset_master<WorkerMaster>` is not that meaningful
```suggestion
    InitializeMaster();
```

src/cpp/benders/benders_by_batch/BendersByBatch.cpp
tbittar:
```suggestion
        _master->addAlphasFixingConstraints(batch_collection_full_for_cuts_);
```

src/cpp/benders/benders_core/include/antares-xpansion/benders/benders_core/WorkerMaster.h
```diff
@@ -4,6 +4,8 @@
 
 #include "SubproblemWorker.h"
 #include "Worker.h"
+#include "antares-xpansion/benders/benders_by_batch/BatchCollection.h"
```
tbittar: I think we need to try removing this dependency...

src/cpp/benders/benders_core/include/antares-xpansion/benders/benders_core/WorkerMaster.h
```diff
@@ -28,6 +30,9 @@ class WorkerMaster: public Worker
     ~WorkerMaster() override = default;
     std::vector<int> _id_master_only_vars;
 
+    void addCutsAlphas(std::vector<SubProblemNamesInCut>&);
```
tbittar:
```suggestion
    void addAlphasFixingConstraints(std::vector<SubProblemNamesInCut>&);
```

src/cpp/benders/benders_core/CMakeLists.txt
```diff
@@ -71,6 +71,7 @@ target_link_libraries(benders_core
         antaresXpansion::outer_loop_lib
         antaresXpansion::core
         antaresXpansion::plugins
+        antaresXpansion::benders_by_batch_core
```
tbittar: Logically the core should not depend on the by_batch

src/cpp/benders/benders_core/WorkerMaster.cpp
```diff
@@ -368,6 +369,52 @@ void WorkerMaster::_set_upper_bounds() const
     _solver->chg_bounds(indices, bndTypes, bounds);
 }
 
+void WorkerMaster::addCutsAlphas(std::vector<SubProblemNamesInCut>& mpi_cuts)
```
tbittar:
```suggestion
void WorkerMaster::addCutsAlphas(std::vector<SubProblemNamesInCut>& names_in_cuts)
```

src/cpp/benders/benders_core/WorkerMaster.cpp
```diff
@@ -368,6 +369,52 @@ void WorkerMaster::_set_upper_bounds() const
...
+{
+    for (auto& cut: mpi_cuts)
```
tbittar:
```suggestion
    for (auto& names_in_cut: names_in_cuts)
```

src/cpp/benders/benders_core/WorkerMaster.cpp
```diff
@@ -368,6 +369,52 @@ void WorkerMaster::_set_upper_bounds() const
...
+    {
+        for (size_t i = 0; i < cut.size() - 1; i++)
+        {
+            size_t start = cut[i].first.find('_') + 1;
```
tbittar: You should rely on the `_name_to_id`  map to find the id of a subproblem from its name and then use `_id_single_subpb_costs_under_approx` to map a subpb id to the id of the corresponding alpha var. This is simpler and most of all more robust (as here you heavily rely on the structure of file name of the sub which we do not control....)

src/cpp/benders/benders_core/WorkerMaster.cpp
```diff
@@ -368,6 +369,52 @@ void WorkerMaster::_set_upper_bounds() const
...
+            size_t end = cut[i].first.find('.', start);
+            auto sub_index_str = cut[i].first.substr(start, end - start);
+            std::stringstream alpha_i;
+            alpha_i << "alpha_" << sub_index_str;
```
tbittar: remove debugging comments if not needed

src/cpp/benders/benders_core/WorkerMaster.cpp
```diff
@@ -368,6 +369,52 @@ void WorkerMaster::_set_upper_bounds() const
...
+            auto alpha_i_pos = _solver->get_col_index(alpha_i.str());
+
+            for (size_t j = i + 1; j < cut.size(); j++)
```
tbittar: You add too many constraints here, we only need alpha_0 = alpha_j for all j (all other are redundant)

tbittar (changes requested):

src/cpp/benders/benders_by_batch/BendersByBatch.cpp
ze-hedi: maybe the bybatch data structure need to be moved to benders_core ?

src/cpp/benders/benders_core/CMakeLists.txt
ze-hedi: yes bvut i need for the by batch data structures

src/cpp/benders/benders_core/WorkerMaster.cpp
ze-hedi: actually it won't work this way. _id_single_subpb_costs_under_approx is just an std::vector<int> so in order to find the right inedx of the sub we need to loop on them in the order. the issue is they are not ordered in the cuts (for example cut 1 can contain sub_1 sub_3 and cut2 can contain sub_0 and sub_2. So i believe it's better to keep it as it is for robustness

src/cpp/benders/benders_core/WorkerMaster.cpp
ze-hedi: there's not couts it's a string stream (a data structure not a print)

src/cpp/benders/benders_by_batch/BendersByBatch.cpp
JasonMarechal25: Instead of:

```
    if (_world.rank() == rank_0)
    {
        _master->addAlphasFixingConstraints(batch_collection_full_for_cuts_);
    }
```

You can do :
```
    if (_world.rank() == rank_0)
    {
        addAlphasFixingConstraints(_master, batch_collection_full_for_cuts_);
    }
```
This way, masterworker doesn't depend on batchCollection.

src/cpp/benders/benders_core/BendersBase.cpp
```diff
@@ -773,6 +773,7 @@ void BendersBase::BuildCutFull(const SubProblemDataMap& subproblem_data_map)
     check_status(subproblem_data_map);
     if (_options.NB_CUTS_PER_ITER)
     {
+        std::cout << "NB_CUTS_PER_ITER " << _options.NB_CUTS_PER_ITER << std::endl;
```
tbittar: I think this is not necessary to print this at each iteration

tbittar (commented):
Good on libs management, still some comments to fix

src/cpp/benders/benders_mpi/BendersMPI.cpp
```diff
@@ -166,6 +166,36 @@ void BendersMpi::BuildMasterProblem()
...
+void BendersMpi::addAlphasFixingConstraints(std::vector<SubProblemNamesInCut>& names_in_cuts)
+{
+    for (auto& names_in_cut: names_in_cuts)
+    {
+        size_t start = names_in_cut[0].first.find('_') + 1;
+        size_t end = names_in_cut[0].first.find('.', start);
+        auto sub_index_str = names_in_cut[0].first.substr(start, end - start);
+        std::stringstream alpha_i;
+        alpha_i << "alpha_" << sub_index_str;
+        auto alpha_i_pos = _master->get_col_index(alpha_i.str());
```
tbittar: Use the maps...

tbittar (changes requested):

src/cpp/benders/benders_core/include/antares-xpansion/benders/benders_core/WorkerMaster.h
```diff
@@ -49,9 +55,9 @@ class WorkerMaster: public Worker
                                const double& rhs) const;
 
     void fix_alpha(const double& bestUB) const;
-
     virtual void DeactivateIntegrityConstraints() const;
     virtual void ActivateIntegrityConstraints() const;
+    void addAlphasFixingConstraints(std::vector<SubProblemNamesInCut>&,std::map<std::string,int>&) ;
```
tbittar: I find it clearer to have variable names here, especially to understand what the map is
```suggestion
    void addAlphasFixingConstraints(std::vector<SubProblemNamesInCut>& names_in_cut,std::map<std::string,int>& problem_to_id) ; 
```

src/cpp/exe/benders/main.cpp
```diff
@@ -10,12 +10,14 @@ int main(int argc, char** argv)
     {
         mpi::environment env(argc, argv);
         mpi::communicator world;
+        std::cout << "in the try of the main of benders " << std::endl;
```
tbittar: Remove debugging comments

src/cpp/exe/benders/main.cpp
```diff
@@ -10,12 +10,14 @@ int main(int argc, char** argv)
...
         // First check usage (options are given)
         if (world.rank() == 0)
         {
             usage(argc);
         }
         auto benders_factory = BendersApp(argv[1], world, SOLVER::BENDERS);
+        std::cout << "benders factory created successfuly" << std::endl;
```
tbittar: Remove debugging comments

tbittar (changes requested):

src/cpp/benders/benders_core/WorkerMaster.cpp
```diff
@@ -383,15 +383,18 @@ int WorkerMaster::get_col_index(std::string variable_id)
...
 }
 
-void WorkerMaster::addAlphasFixingConstraints(std::vector<SubProblemNamesInCut>& names_in_cuts,std::map<std::string,int>& problem_to_id) 
+void WorkerMaster::addAlphasFixingConstraints(std::vector<SubProblemNamesInCut>& names_in_cuts,
+                                              std::map<std::string, int>& problem_to_id)
 {
     for (auto& names_in_cut: names_in_cuts)
     {
-        auto alpha_i_pos = _id_single_subpb_costs_under_approx[ problem_to_id[names_in_cut[0].first]] ; 
+        auto alpha_i_pos = _id_single_subpb_costs_under_approx
```
tbittar: Rename `alpha_0_pos`

tbittar (approved):

PR #1222, opened 2026-03-24
Adjust cut coefficient tolerance to weights

Juliette-Gerbaux:


src/cpp/benders/benders_core/WorkerMaster.cpp
```diff
@@ -516,3 +518,22 @@ void WorkerMaster::ActivateIntegrityConstraints() const
...
+                                          int first,
+                                          int last,
+                                          const std::vector<int>& subproblem_ids) const
+{
+    double cut_coefficient_tolerance = 0.0;
+    for (int sb_id: subproblem_ids)
+    {
+        cut_coefficient_tolerance += _subproblem_tolerance.find(sb_id)->second;
+    }
+
```
tbittar, line 578: Ideally we could avoid to repeat this computation at each iteration as it will be the same for each identical cut. This could be done after cut definitions in the initialization of workers. (but this have not major impact of perfs I guess as the computation is really light so you choose whether or not to take into account this remark based on the implementation difficulty

tests/cpp/benders/WorkerMasterTest.cpp
```diff
@@ -39,6 +39,23 @@ class NOOPSolverForWorkerMaster: public NOOPSolverForWorker
...
+              const double* dmatval,
+              const std::vector<std::string>& row_names) override
+    {
+        last_rowrhs.assign(rhs, rhs + newrows);
+        last_matval.assign(dmatval, dmatval + newnz);
+    }
+
+    std::vector<double> last_rowrhs;
+    std::vector<double> last_matval;
+
```
tbittar: Almost same feature as the ongoing work in https://github.com/AntaresSimulatorTeam/antares-xpansion/pull/1215, we should synchronize to avoid future conflict. I prefer the other PRs approach using a derived class because adding an implementation directly in NOOP does not make it not operational anymore. Can you check if the other design adapts easily to your case ?

src/cpp/benders/benders_core/BendersBase.cpp
```diff
@@ -1497,11 +1497,14 @@ void BendersBase::SetPlugin(std::shared_ptr<BendersPlugin> benders_plugin)
 std::map<int, double> BendersBase::GetSubCutTolerance() const
 {
     std::map<int, double> subproblem_cut_coefficient_tolerance{};
-    for (const auto& subproblem_name: GetSubProblemNames())
+    for (const auto& subproblem: _problem_to_id)
     {
-        subproblem_cut_coefficient_tolerance[ProblemToId(subproblem_name)]
-          = Options().CUT_COEFFICIENT_TOLERANCE
-            * SubproblemWeight(_data.nsubproblem, subproblem_name);
+        std::cout << subproblem.first << std::endl;
```
tbittar: remove debug print

src/cpp/benders/benders_core/WorkerMaster.cpp
```diff
@@ -519,19 +560,29 @@ void WorkerMaster::ActivateIntegrityConstraints() const
...
+                                              const std::vector<int>& subproblem_ids) const
 {
     double cut_coefficient_tolerance = 0.0;
     for (int sb_id: subproblem_ids)
     {
-        cut_coefficient_tolerance += _subproblem_tolerance.find(sb_id)->second;
+        auto it = _subproblem_tolerance.find(sb_id);
+        if (it == _subproblem_tolerance.end())
+        {
+            throw std::runtime_error("sb_id absent de _subproblem_tolerance");
```
tbittar: Use English for error message

src/cpp/benders/benders_core/WorkerMaster.cpp
```diff
@@ -519,19 +560,29 @@ void WorkerMaster::ActivateIntegrityConstraints() const
...
+        }
+        cut_coefficient_tolerance += it->second;
     }
 
     const double tolerance = cut_coefficient_tolerance;
 
+    if (tolerance >= 1)
+    {
+        throw std::runtime_error("Cut tolerance greater than 1");
```
tbittar: Why ? If subproblem have very small weight it can happen

tests/cpp/benders/WorkerMasterTest.cpp
```diff
@@ -237,11 +359,13 @@ TEST_F(WorkerMasterTest, SetMasterOnlyVarIdsLogic)
     EXPECT_EQ(master->_id_master_only_vars, expected);
 }
 
-TEST_F(WorkerMasterTest, AddSubproblemCutAppliesRoundingOnCoeffs)
+TEST_F(WorkerMasterAlphasFixingTest, AddSubproblemCutAppliesRoundingOnCoeffs)
```
tbittar: The class should be renamed as it covers a larger scope than that maybe `WorkerMasterAddRowsTest`

tbittar (changes requested):

tests/cpp/benders/WorkerMasterTest.cpp
```diff
@@ -137,6 +139,10 @@ class WorkerMasterAlphasFixingTest : public ::testing::Test
 
     std::shared_ptr<WorkerMaster> make_master(int subproblems_count)
     {
```
tbittar, line 141: Bug: the loop always assigns to key `0` instead of `i`, so the map has a single entry regardless of `subproblems_count`. Should be:
```cpp
subproblem_cut_coefficient_tolerance[i] = 0.1;
```

src/cpp/benders/benders_core/BendersBase.cpp
```diff
@@ -1495,3 +1493,18 @@ void BendersBase::SetPlugin(std::shared_ptr<BendersPlugin> benders_plugin)
 {
     benders_plugin_ = benders_plugin;
 }
+
+std::map<int, double> BendersBase::GetSubCutTolerance() const
+{
+    std::map<int, double> subproblem_cut_coefficient_tolerance{};
```
tbittar, line 1499: `Options()` creates a default-constructed object — `_options.CUT_COEFFICIENT_TOLERANCE` (the actual configured value) is never used here. Should this be `_options.CUT_COEFFICIENT_TOLERANCE`? The same pattern exists in BendersMPI/Sequential but may be a latent bug in all three places.

src/cpp/benders/benders_core/WorkerMaster.cpp
```diff
@@ -557,3 +559,32 @@ void WorkerMaster::ActivateIntegrityConstraints() const
     std::vector<char> col_types(_id_int_vars.size(), 'I');
     _solver->chg_col_type(_id_int_vars, col_types);
 }
+
+void WorkerMaster::setToZeroIfWithinTolerance(std::vector<double>& values,
+                                              int first,
+                                              int last,
+                                              const std::vector<int>& subproblem_ids) const
+{
+    double cut_coefficient_tolerance = 0.0;
```
tbittar, line 568: The local variable `tolerance` is just a `const` copy of `cut_coefficient_tolerance` declared 4 lines above. Remove one of the two — either accumulate directly into `tolerance`, or drop the alias and use `cut_coefficient_tolerance` in the lambda.

tbittar (changes requested):
## Review

### Issues requiring fixes

Some already flagged, adding detail and new findings:

1. **Debug print** (`BendersBase.cpp:1502`) — already flagged, must be removed.
2. **French error message** (`WorkerMaster.cpp:574`) — already flagged, use English.
3. **`tolerance >= 1` guard** (`WorkerMaster.cpp:583`) — already questioned. With non-uniform weights or a large configured `CUT_COEFFICIENT_TOLERANCE`, the sum can legitimately reach ≥ 1. Either document why this is a hard invariant or remove the guard.
4. **Bug in test fixture** (`WorkerMasterAlphasFixingTest::make_master`, line ~141): the loop always writes to key `0` instead of key `i`, so the map always has a single entry regardless of `subproblems_count`. Should be `subproblem_cut_coefficient_tolerance[i] = 0.1;`.
5. **`Options()` vs `_options`** in `GetSubCutTolerance`: creates a default-constructed object, so any user-configured `CUT_COEFFICIENT_TOLERANCE` is silently ignored. The same pattern exists in BendersMPI/Sequential — may be a latent bug in all three places.
6. **Redundant alias** in `setToZeroIfWithinTolerance`: `cut_coefficient_tolerance` is summed then immediately copied into `const double tolerance`. One variable should be removed.
7. **Test class name** (`WorkerMasterAlphasFixingTest`) — already flagged for renaming. The new `AddSubproblemCutAppliesRoundingOnCoeffs` test has nothing to do with alpha-fixing.

### Positive notes
- Moving rounding to the master where both RHS and matval are in scope is cleaner than the previous approach.
- Scaling the tolerance by subproblem weight is the right approach for non-uniform configurations.
- Removing `SubproblemWorkerTest.cpp` and replacing it with new `WorkerMasterTest` coverage is correct.

src/cpp/benders/benders_core/WorkerMaster.cpp
Juliette-Gerbaux: In the cut there is the coefficient for the alpha variable which is -1 so we don't want to remove this coefficient. Maybe I could add some precisions.

src/cpp/benders/benders_core/BendersBase.cpp
Juliette-Gerbaux: Options() returns _options

tbittar (approved):

PR #1225, opened 2026-03-25
feat: Test if gemsinvest workflows succeed

payetvin:


tests/end_to_end/gems/test_gemsEndToEnd.py
```diff
@@ -0,0 +1,105 @@
...
+        output_data = json.load(f)
+
+    assert "solution" in output_data, "No solution in out.json"
+    assert "problem_status" in output_data["solution"], "No problem_status in solution"
+    assert output_data["solution"]["problem_status"] == "OPTIMAL", \
+        f"Expected OPTIMAL status, got {output_data['solution']['problem_status']}"
+
+    solution = output_data["solution"]
+
+    if "overall_cost" in expected_results:
```
tbittar: As we provide the expected_results file I do not see why we need `if`, especially, if it is not here, we do not test anything and the test may succeed without doing anything

tests/end_to_end/gems/test_gemsEndToEnd.py
```diff
@@ -0,0 +1,105 @@
...
+        actual_cost = solution.get("overall_cost")
+        expected_cost = expected_results["overall_cost"]
+        assert actual_cost is not None, "No overall_cost in solution"
+        np.testing.assert_allclose(actual_cost, expected_cost, rtol=1e-6, atol=0,
+                                    err_msg=f"Overall cost mismatch: expected {expected_cost}, got {actual_cost}")
+
+    if "variables" in expected_results:
```
tbittar: Same remark

tests/CMakeLists.txt
```diff
@@ -188,6 +188,17 @@ if (PYTHON_MODULE_pytest_FOUND AND PYTHON_MODULE_numpy_FOUND)
     )
     set_property(TEST merge_mps PROPERTY LABELS benders merge-mps end_to_end)
 
+    # gems end to end tests
+    add_test(
+            NAME gems_sequential
```
tbittar:
```suggestion
            NAME e2e_gems
```

data_test/simulator_hybrid_invest_13_1/expected_results.json
tbittar, line 1: I would rather store this within the study (as it is done like this for other studies), also if we add more studies, we may not want having too many "expected_results" file in this dir

tests/end_to_end/gems/conftest.py
```diff
@@ -0,0 +1,23 @@
+import pytest
+import sys
+from pathlib import Path
+
+sys.path.insert(0, str(Path(__file__).parent.parent.parent.parent / "src" / "python"))
```
tbittar: Why do we this and not in https://github.com/AntaresSimulatorTeam/antares-xpansion/blob/37cb28df1fd4bd3150b270bf12aa6e0fdd8490ff/tests/end_to_end/examples/conftest.py for instance ?

tbittar (changes requested):

tbittar (commented):
Now that full GEMS workflow suceeds, you could even add test full modeler 13_1 (the expected results should be the same as for the one added in this PR)

tbittar (approved):

flomnes (commented):
Use 168 timesteps for the full GEMS study, check that the results are identical vs hybrid

tbittar (approved):

PR #1227, opened 2026-03-27
Base for gemsinvest in full gems

payetvin:


payetvin:
Not needed, allowing antares-problem-generator to load GEMS studies is simpler

PR #1229, opened 2026-04-01
Ignore candidate in GEMS workflow, allow full GEMS studies

payetvin:


src/python/antares_xpansion/config_loader.py
```diff
@@ -373,7 +396,11 @@ def __init__(self, config: XpansionConfig):
         :param config: configuration to use for the optimization
         :type config: XpansionConfig object
         """
-        super().__init__(config.data_dir, config)
+        skip_file_validation = (
+            config.step in ("benders", "full", "gems")
+            and XpansionSettingsReader._has_optim_config_static(config.data_dir, config)
```
tbittar: Why can't you use `has_optim_config` directly and need a static version ? Or even remove the attribute, see https://github.com/AntaresSimulatorTeam/antares-xpansion/pull/1229#discussion_r3080489981

src/python/antares_xpansion/driver.py
```diff
@@ -110,12 +110,11 @@ def launch(self):
         """
 
         use_gems = (
-            self.config_loader.step() in ("benders", "full")
+            self.config_loader.step() in ("gems", "benders", "full")
```
tbittar: Step `benders` is meant to launch the benders executable on a folder with all problems already generated + a structure file (no matter how, from gems study or not), I would not include `benders` here as this step is not impacted by the type of input study
```suggestion
            self.config_loader.step() in ("gems", "full")
```

src/python/antares_xpansion/gems_driver.py
```diff
@@ -84,8 +85,13 @@ def antares_output_dir(self):
         """
         return os.path.normpath(os.path.join(self.data_dir, self.output))
 
-    def _set_simulation_name(self):
+    def _set_simulation_name(self, study_is_full_gems: bool):
         self.simulation_name = ""
+
+        # files are directly in the output/ dir for full GEMS, skip the search for the latest dir
+        if study_is_full_gems: 
+            return;
```
tbittar: The comma after `return` is intended ?

src/python/antares_xpansion/driver.py
```diff
@@ -110,12 +110,11 @@ def launch(self):
         """
 
         use_gems = (
```
tbittar: Add a comment that says that this case handles:
- Full GEMS studies
- Hybrid studies with GEMS investment candidates

Could be renamed `use_gems_candidates` (as hybrid studies with legacy candidates cannot be handled with the gems_step -> they need the legacy Xpansion problem generator -> not handled)

src/python/antares_xpansion/driver.py
```diff
@@ -110,12 +110,11 @@ def launch(self):
...
-            self.config_loader.step() in ("benders", "full")
+            self.config_loader.step() in ("gems", "benders", "full")
             and self.config_loader.has_optim_config()
         )
 
         if use_gems:
```
tbittar: It seems that the block L178 is then never reached (and may be deleted) ?

src/python/antares_xpansion/config_loader.py
```diff
@@ -146,6 +167,8 @@ def check_candidates_file_format(self):
         )
 
     def check_settings_file_format(self):
+        if self._skip_file_validation:
```
tbittar: Even with full GEMS studies we rely on the settings file for Benders settings (for now), so these checks are still necessary.

Except that additionnal_constraints will not be taken into account (as they use the legacy problem-generator) so raise an error within `self._verify_additional_constraints_file()` if we have an additionnal_constraint setting when using GEMS

src/python/antares_xpansion/config_loader.py
```diff
@@ -40,10 +40,12 @@ class XpansionSettingsReader:
     Minimal class to read data in a study's setting
     """
 
-    def __init__(self, study_path: Path, xpansion_defaults: XpansionConfigConstants):
+    def __init__(self, study_path: Path, xpansion_defaults: XpansionConfigConstants,
+                 skip_file_validation: bool = False):
```
tbittar: Only the candidate file check needs to be skipped when using GEMS candidates, clearer to understand in what situation we arrive to skip the check when reading the code
```suggestion
                 use_gems_candidates: bool = False):
```

src/python/antares_xpansion/config_loader.py
```diff
@@ -115,6 +117,23 @@ def has_optim_config(self):
...
+        optim_config_path = os.path.normpath(
+            os.path.join(
+                data_dir,
+                config.INPUT,
+                "optim-config.yml",
+            )
+        )
+        return os.path.isfile(optim_config_path)
+
+    def is_gems_mode(self):
```
tbittar:
```suggestion
    def use_gems_candidates(self):
```

src/python/antares_xpansion/driver.py
tbittar: And why not simply use `self.config_loader.is_gems_mode()` ?

src/python/antares_xpansion/config_loader.py
```diff
@@ -115,6 +117,23 @@ def has_optim_config(self):
...
+        """
+        Check if gems workflow should be used (gems step or benders/full step with optim-config.yml)
+        """
+        return self.step() in ("gems", "benders", "full") and self.has_optim_config()
```
tbittar:
```suggestion
        return self.step() in ("gems", "full") and self.has_optim_config()
```

src/python/antares_xpansion/config_loader.py
tbittar: Or is it even possible to remove this attribute and simply do 
```python
if not self.is_gems_mode(): 
    self.check_candidates_file_format()```

**Review by tbittar — CHANGES_REQUESTED** [review]

File: src/python/antares_xpansion/config_loader.py [hunk 10]
```diff
@@ -117,22 +114,8 @@ def has_optim_config(self):
...
-        )
-        return os.path.isfile(optim_config_path)
-
-    def is_gems_mode(self):
-        """
-        Check if gems workflow should be used (gems step or benders/full step with optim-config.yml)
-        """
-        return self.step() in ("gems", "benders", "full") and self.has_optim_config()
+    def gems_candidates(self):
+        return self._gems_candidates
```
tbittar, line 118: Useful ?

src/python/antares_xpansion/config_loader.py
```diff
@@ -40,17 +40,17 @@ class XpansionSettingsReader:
...
 
-    def __init__(self, study_path: Path, xpansion_defaults: XpansionConfigConstants,
-                 skip_file_validation: bool = False):
+    def __init__(self, study_path: Path, xpansion_defaults: XpansionConfigConstants):
         self.path = study_path
         self._config_defaults = xpansion_defaults
         self.logger = step_logger(__name__, __class__.__name__)
-        self._skip_file_validation = skip_file_validation
 
+        self._gems_candidates = self._check_gems_candidates()
```
tbittar: Rename suggestion : `self.gems_candidates = self._use_gems_candidates()`

tbittar (approved):

src/python/antares_xpansion/config_loader.py
payetvin: The class has methods to access it's members like this:
```
    def step(self):
        return self._config.step

    def simulation_name(self):
        return self._xpansion_simulation_name
```
I kept the same design, we can keep only _gems_candidates it's fine for me

tbittar (approved):

PR #1250, opened 2026-06-05
memory optimization version 1

ze-hedi:
Memory optimization new algo : 

Cloning skeleton object for each subproblem during subproblem solving


src/cpp/benders/benders_core/include/antares-xpansion/benders/benders_core/BendersBase.h
```diff
@@ -14,6 +14,7 @@
 #include "SubproblemWorker.h"
 #include "Worker.h"
 #include "WorkerMaster.h"
+#include "antares-xpansion/benders/benders_core/MemOptim_subproblem_builder.h"
```
tbittar: Why not the following ?
```suggestion
#include "MemOptim_subproblem_builder.h"
```

src/cpp/benders/benders_core/BendersBase.cpp
```diff
@@ -482,12 +482,21 @@ void BendersBase::GetSubproblemCut(SubProblemDataMap& subproblem_data_map)
     {
         GetSubproblemCutCache(subproblem_data_map);
     }
+    else if (Options().MEMORY_OPTIMIZATION)
```
tbittar: As `CACHE_PROBLEMS` and `MEMORY_OPTIMIZATION` are mutually exclusive, why not having a single option `CACHE_PROBLEMS` that can take 3 values (with explicit names) : 
- `none` (equivalent of cache_problems = False, memory_optimization = False)
- `on_disk` (equivalent of cache_problems=true, memory_optimization = False)
- `rhs_and_obj_in_memory` (equivalent of cache_problems=False, memory_optimization=True)

this is more robust as we avoid the possibility of having cache_problems=true and memory_optimization=True which is not well handled as `BendersMpi::InitializeProblems` checks memory_optimization first and `BendersBase::GetSubproblemCuts` checks cache_problems first which could cause a pb if both options are true

src/cpp/benders/benders_core/BendersBase.cpp
```diff
@@ -619,6 +628,26 @@ void BendersBase::GetSubproblemCutCache(SubProblemDataMap& subproblem_data_map)
...
+    {
+        auto variable_map = coupling_map_[sub];
+        double slave_weights = SubproblemWeight(memoptim_subprob_builder_->get_sub_number(), sub);
+        auto subproblem_worker = memoptim_subprob_builder_->create_sub_solver_abstract(
+          sub,
+          variable_map,
+          _options.CUT_COEFFICIENT_TOLERANCE,
+          slave_weights);
+        PlainData::SubProblemData subproblem_data;
+        SolveSubproblem(subproblem_data, sub, subproblem_worker);
```
tbittar, line 647: Here implicitly, you reuse the basis of the previous subproblem. Is it worth to reuse the basis from the previous resolution of the same sub ? What is the interaction with micro iterations ?

src/cpp/benders/benders_core/include/antares-xpansion/benders/benders_core/MemOptim_subproblem_builder.h
```diff
@@ -0,0 +1,66 @@
...
+
+class MemOptimSubProblemBuilder
+{
+public:
+    MemOptimSubProblemBuilder(const std::filesystem::path& inputRoot,
+                              Logger& logger,
+                              std::string solver_name,
+                              int log_level,
+                              ProblemsFormat format);
+    void build_sub_skeleton(std::string solver_name,
```
tbittar: Could be private ?

src/cpp/benders/benders_core/MemOptim_subproblem_builder.cpp
```diff
@@ -0,0 +1,270 @@
...
+    micro_iters_ = false;
+    warm_start_ = false;
+}
+
+void MemOptimSubProblemBuilder::read_coef()
+{
+    boost::char_separator<char> sep(",");
+    using Tokenizer = boost::tokenizer<boost::char_separator<char>>;
+
+    auto coef_csv_path = inputRoot_ / "sub" / "coef.csv";
```
tbittar: We can keep the data structure hard coded for now, but then we need to add a part in the docs that describe the expected study structure when using the memory_optimization feature

tbittar: And describing the expected format of each csv file (by the way is it useful to save disk space to generate these data in a more compact form like parquet ?)

src/cpp/benders/benders_core/MemOptim_subproblem_builder.cpp
```diff
@@ -0,0 +1,270 @@
...
+    std::ifstream coef_csv_stream(coef_csv_path);
+    if (coef_csv_stream.is_open())
+    {
+        std::string line;
+        while (std::getline(coef_csv_stream, line))
+        {
+            Tokenizer tok(line, sep);
+            std::vector<std::string> tokens(tok.begin(), tok.end());
+            std::string key = tokens[0];
```
tbittar: Use a more explicit name for key : `sub_name` ?

src/cpp/benders/benders_core/MemOptim_subproblem_builder.cpp
```diff
@@ -0,0 +1,270 @@
...
+#include <chrono>
+#include <fstream>
+#include <iostream>
+
+#include <boost/mpi.hpp>
+#include <boost/tokenizer.hpp>
+
+#include "antares-xpansion/benders/benders_core/SubproblemWorker.h"
+
+MemOptimSubProblemBuilder::MemOptimSubProblemBuilder(const std::filesystem::path& inputRoot,
```
tbittar: I think it could be very valuable to have some unit test with a very small csvs (say 3/4 lines) to unit test all the `read` functions. this will also allow to have examples of the expected format

src/cpp/benders/benders_core/MemOptim_subproblem_builder.cpp
```diff
@@ -0,0 +1,270 @@
...
+    read_coef_rows();
+    read_obj_coef();
+    read_obj_cols();
+    read_rhs();
+    read_rhs_rows();
+    micro_iters_ = false;
+    warm_start_ = false;
+}
+
+void MemOptimSubProblemBuilder::read_coef()
```
tbittar: Maybe add a small comment at the beginning of each `read` function to explain the expected columns of the csv

src/cpp/benders/benders_core/MemOptim_subproblem_builder.cpp
```diff
@@ -0,0 +1,270 @@
...
+{
+    boost::char_separator<char> sep(",");
+    using Tokenizer = boost::tokenizer<boost::char_separator<char>>;
+
+    auto obj_coef_path = inputRoot_ / "sub" / "obj_coef.csv";
+    std::ifstream obj_coef_stream(obj_coef_path);
+    if (obj_coef_stream.is_open())
+    {
+        std::string line;
+        int j = 0;
```
tbittar: Useless variable ?

src/cpp/benders/benders_core/MemOptim_subproblem_builder.cpp
```diff
@@ -0,0 +1,270 @@
...
+{
+    logger_ = logger;
+    build_sub_skeleton(solver_name, solver_log_manager_, log_level, format);
+    read_coef();
+    read_coef_cols();
+    read_coef_rows();
+    read_obj_coef();
+    read_obj_cols();
+    read_rhs();
+    read_rhs_rows();
```
tbittar: i think you could refactor in only 2 functions to remove duplicated code : `read_coef(coef_type)` with `coef_type` being either matrix, obj or rhs to get the `coefs` attribute and one`read_name_and_index(row | col, obj | rhs | matrix_coef)`

src/cpp/benders/benders_core/include/antares-xpansion/benders/benders_core/MemOptim_subproblem_builder.h
```diff
@@ -0,0 +1,66 @@
...
+    void read_obj_cols();
+    void read_rhs();
+    void read_rhs_rows();
+
+    std::filesystem::path inputRoot_;
+
+    std::map<std::string, std::vector<double>> coeffs_;
+    std::map<std::string, std::vector<SolverRepresentedRows>> micro_iters_added_rows;
+    std::vector<std::string> coef_cols_;
+    std::vector<std::string> coef_rows_;
```
tbittar: Do we need to store this as class atrtibute as it seems it is only used as intermediary data in `read_coef_cols()` and `read_coef_rows()` to build `constraints_cols_indices` and `constraints_cols_indices_` ? Same question for `obj_cols` and `rhs_rows_`

src/cpp/benders/benders_core/FixedSkeletonSubProblemBuilder.cpp
```diff
@@ -0,0 +1,270 @@
...
+                                                   ProblemsFormat format)
+{
+    SolverFactory solver_factory(logger_);
+    solver_ = solver_factory.create_solver(solver_name,
+                                           SOLVER_TYPE::CONTINUOUS,
+                                           solver_log_manager);
+
+    solver_->set_threads(1);
+    solver_->set_output_log_level(log_level);
+    std::filesystem::path skeleton_sub = inputRoot_ / "sub" / "sub.mps";
```
tbittar, line 140: Indicate clearly in the future docs on the expected format that only mps is supported for now for the skeleton

src/cpp/benders/benders_core/MemOptim_subproblem_builder.cpp
```diff
@@ -0,0 +1,270 @@
...
+}
+
+std::shared_ptr<SubproblemWorker> MemOptimSubProblemBuilder::create_sub_solver_abstract(
+  std::string sub_name,
+  VariableMap& variable_map,
+  double cut_coefficient_tolerance,
+  double slave_weight)
+{
+    auto& coeffs_sub = coeffs_[sub_name];
+    auto& coeffs_obj = obj_coefs_[sub_name];
```
tbittar: for consistency
```suggestion
    auto& coeffs_obj = obj_coeffs_[sub_name];
```

src/cpp/benders/benders_core/MemOptim_subproblem_builder.cpp
```diff
@@ -0,0 +1,270 @@
...
+    auto& rhs_values = rhs_[sub_name];
+    int n_coefs = coeffs_sub.size();
```
tbittar: for consistency
```suggestion
    int n_coeffs = coeffs_sub.size();
```

src/cpp/benders/benders_core/MemOptim_subproblem_builder.cpp
```diff
@@ -0,0 +1,270 @@
...
+    std::shared_ptr<SolverAbstract> sub_solver(solver_->clone());
+
+    auto start = std::chrono::high_resolution_clock::now();
```
tbittar: Used for testing, remove in production code

src/cpp/benders/benders_core/MemOptim_subproblem_builder.cpp
```diff
@@ -0,0 +1,270 @@
...
+    sub_solver->chg_coefs(n_coefs,
+                          constraints_row_indices_.data(),
+                          constraints_col_indices_.data(),
+                          coeffs_sub.data());
+    sub_solver->chg_obj(obj_col_indices_, coeffs_obj);
+    sub_solver->chg_rhs_values(rhs_row_indices_, rhs_values);
+    auto end = std::chrono::high_resolution_clock::now();
+    auto duration = std::chrono::duration_cast<std::chrono::microseconds>(end - start);
+
```
tbittar: Used for testing, remove for production code

src/cpp/benders/benders_core/MemOptim_subproblem_builder.cpp
```diff
@@ -0,0 +1,270 @@
...
+}
+
+void MemOptimSubProblemBuilder::read_coef_cols()
+{
+    boost::char_separator<char> sep(",");
+    using Tokenizer = boost::tokenizer<boost::char_separator<char>>;
+
+    auto coef_cols_path = inputRoot_ / "sub" / "coef_cols.csv";
+    std::ifstream coef_cols_stream(coef_cols_path);
+    if (coef_cols_stream.is_open())
```
tbittar: Raise an error if error on opening, same for all other read functions

src/cpp/benders/benders_core/FixedSkeletonSubProblemBuilder.cpp
```diff
@@ -0,0 +1,270 @@
...
+
+int MemOptimSubProblemBuilder::get_sub_number()
+{
+    return rhs_.size();
+}
+
+std::shared_ptr<SubproblemWorker> MemOptimSubProblemBuilder::create_sub_solver_abstract(
+  std::string sub_name,
+  VariableMap& variable_map,
+  double cut_coefficient_tolerance,
```
tbittar, line 156: Unused ?

src/cpp/benders/benders_core/SubproblemWorker.cpp
```diff
@@ -38,6 +38,33 @@ SubproblemWorker::SubproblemWorker(const VariableMap& variable_map,
...
+                                   double slave_weight):
+    Worker(variable_map, logger)
+{
+    init_for_mem_optim(solver, variable_map);
+    setup_obj(slave_weight);
+}
+
+void SubproblemWorker::setup_obj(double slave_weight)
+{
+    int mps_ncols(_solver->get_ncols());
```
tbittar, line 53: mps has nothing to do here (although this pattern is used elsewhere in the code..)
```suggestion
    int ncols(_solver->get_ncols());
```

src/cpp/benders/benders_core/SubproblemWorker.cpp
```diff
@@ -38,6 +38,33 @@ SubproblemWorker::SubproblemWorker(const VariableMap& variable_map,
...
+                                   std::shared_ptr<SolverAbstract> solver,
+                                   Logger logger,
+                                   double slave_weight):
+    Worker(variable_map, logger)
+{
+    init_for_mem_optim(solver, variable_map);
+    setup_obj(slave_weight);
+}
+
+void SubproblemWorker::setup_obj(double slave_weight)
```
tbittar, line 51: Could we also call `setup_obj` in the normal `SubproblemWorker` initialization to avoid code duplication ?

src/cpp/benders/benders_core/Worker.cpp
```diff
@@ -125,6 +125,21 @@ void Worker::solve(int& lp_status,
     }
 }
 
+void Worker::set_id_to_name(VariableMap& variable_map)
```
tbittar, line 128: You can call it within `Worker::init()` to avoid code duplication ?

src/cpp/benders/benders_core/Worker.cpp
```diff
@@ -125,6 +125,21 @@ void Worker::solve(int& lp_status,
...
+{
+    _name_to_id = variable_map;
```
tbittar, line 130: this seems to be done by the `Worker` constructor, do we need to update it ?

src/cpp/benders/benders_mpi/BendersMPI.cpp
```diff
@@ -34,40 +36,64 @@ void BendersMpi::InitializeProblems()
 {
     MatchProblemToId();
     SubProblemNamesInCut subs_per_proc;
-    if (_options.CACHE_PROBLEMS)
-    {
+    if (_options.MEMORY_OPTIMIZATION)
```
tbittar: Is bendersbyBatch compatible with mem optim ?

src/cpp/lpnamer/model/include/antares-xpansion/lpnamer/model/Problem.h
```diff
@@ -278,6 +278,16 @@ class Problem: public SolverAbstract
         solver_abstract_->chg_coef(id_row, id_col, val);
     }
 
+    void chg_rhs_values(std::vector<int>& id_cols, std::vector<double>& vals)
+    {
+        solver_abstract_->chg_rhs_values(id_cols, vals);
+    }
+
+    void chg_coefs(int num_coefs, int* id_rows, int* id_cols, double* vals)
```
tbittar: Why do you have a `std::vector<int>` for `id_cols` in `chg_rhs_values` and a `int*` here ? Is it possible to uniformize ?

src/cpp/multisolver_interface/private/SolverClp.h
```diff
@@ -156,6 +156,8 @@ class SolverClp: public SolverAbstract
     void chg_col_type(const std::vector<int>& mindex, const std::vector<char>& qctype) override;
     void chg_rhs(int id_row, double val) override;
     void chg_coef(int id_row, int id_col, double val) override;
+    void chg_coefs(int num_coefs, int* id_rows, int* id_cols, double* vals);
+    void chg_rhs_values(std::vector<int>&, std::vector<double>&);
```
tbittar: Missing override

src/cpp/multisolver_interface/SolverCbc.cpp
```diff
@@ -713,6 +713,10 @@ void SolverCbc::chg_rhs(int id_row, double val)
     }
 }
 
+void SolverCbc::chg_coefs(int num_coefs, int* id_rows, int* id_cols, double* vals)
```
tbittar: Please implement this, even by doing a for loop over `chg_coef`. The goal is not to be efficient but at least avoiding a silent wrong behavior where coefficients are not updated without raising any errors or warning. Same for `chg_rhs_values` and `SolverClp`

tbittar (changes requested):

src/cpp/benders/benders_core/BendersBase.cpp
tbittar: Impact on this change :
- End to end tests to update
- antares-xpansion-launcher command line option `--cache-problems` should be updated to handle the new format and handle the changes in the Python part of the drivers (options file generation)
- Documentation update

src/cpp/benders/benders_core/Worker.cpp
ze-hedi: it's used in init_for_mem_optim which is called in subProblemWorker constructor that is called on MemOptim case

src/cpp/benders/benders_core/ConstraintsFileReader.cpp
```diff
@@ -12,12 +12,27 @@ ConstraintsFileReader::ConstraintsFileReader(const std::filesystem::path constra
     solver_ = solver_factory.create_solver(solver_name,
                                            SOLVER_TYPE::CONTINUOUS,
                                            solver_log_manager);
-    solver_->set_threads(1);
-    solver_->set_output_log_level(log_level);
+
+    if (solver_ != nullptr)
```
tbittar: Is this not handled in the `create_solver` method ?

src/cpp/benders/benders_core/ConstraintsFileReader.cpp
```diff
@@ -12,12 +12,27 @@ ConstraintsFileReader::ConstraintsFileReader(const std::filesystem::path constra
...
+    }
+    else
+    {
+        std::cout << "solver is null " << std::endl;
+    }
 
     benders_problem_provider_ = std::make_shared<BendersProblemFromFile>(constraint_file_path);
     solver_IO_.configure(solver_name, format);
-    benders_problem_provider_->provide_problem(solver_IO_, solver_);
+    if (benders_problem_provider_ != nullptr)
```
tbittar: same question

data_test/low_memory_study/sub/.~lock.coef.csv#
tbittar, line 1: remove

src/cpp/benders/benders_core/SimulationOptions.cpp
```diff
@@ -203,7 +202,7 @@ BendersBaseOptions SimulationOptions::get_benders_options() const
     result.BOUND_ALPHA = BOUND_ALPHA;
     result.CACHE_PROBLEMS = CACHE_PROBLEMS;
 
-    std::cout<<"CACHE_PROBLEMS "<<CACHE_PROBLEMS<<std::endl ; 
+    std::cout << "CACHE_PROBLEMS " << CACHE_PROBLEMS << std::endl;
```
tbittar, line 205: Remove

src/cpp/benders/benders_mpi/BendersMPI.cpp
```diff
@@ -38,67 +38,66 @@ void BendersMpi::InitializeProblems()
     SubProblemNamesInCut subs_per_proc;
     switch (_options.CACHE_PROBLEMS)
     {
-        case 2:
+    case 2:
+    {
+        std::cout << "case mem optim " << std::endl;
```
tbittar: remove

tbittar (changes requested):

src/cpp/benders/benders_core/FixedSkeletonSubProblemBuilder.cpp
tbittar: Remove this unused argument

src/cpp/benders/benders_core/include/antares-xpansion/benders/benders_core/MemOptim_subproblem_builder.h
```diff
@@ -0,0 +1,76 @@
...
+#include "antares-xpansion/xpansion_interfaces/ILogger.h"
+
+enum class CoeffType
+{
+    constraints = 0,
+    objective,
+    rhs
+};
+
+class MemOptimSubProblemBuilder
```
tbittar:
```suggestion
class SubProblemFromSkeletonPlusRhsBuilder
```

tests/cpp/benders/MemOptimSubProblemBuilderTest.cpp
```diff
@@ -0,0 +1,197 @@
...
+#include <string>
+#include <vector>
+
+#include <gtest/gtest.h>
+
+#include "LoggerStub.h"
+#include "NOOPSolver.h"
+#include "antares-xpansion/benders/benders_core/MemOptim_subproblem_builder.h"
+
+class MemOptimSubProblemBuilderTest: public ::testing::Test
```
tbittar: At least some tests should verify the content of the created solver abstracts and see if it matches the data in the CSVs

src/cpp/benders/benders_mpi/BendersMPI.cpp
tbittar: Please fix this !

src/cpp/benders/benders_core/Worker.cpp
tbittar: set_id_to_name has exactly one call site — inside apply_per_subproblem_varying_coefficients — and nowhere else. It's a two-liner that exists only as an extracted helper for that one function. It could be inlined directly into apply_per_subproblem_varying_coefficients and removed entirely:


void Worker::apply_per_subproblem_varying_coefficients(
    std::shared_ptr<SolverAbstract> solver, VariableMap& variable_map)
{
    _solver = solver;
    _name_to_id = variable_map;
    for (const auto& kvp: _name_to_id)
        _id_to_name[kvp.second] = kvp.first;
}
It is also declared public in Worker.h despite never being called from outside the class, which makes the interface noisier than it needs to be. Unless you anticipate other callers, the answer is no — it has no independent value and can be collapsed.

tbittar (changes requested):

tbittar:
You need to update the docs in `docs/user-guide/get-started/options-definition.md` for cache_problems

src/cpp/benders/benders_core/include/antares-xpansion/benders/benders_core/FixedSkeletonSubProblemBuilder.h
```diff
@@ -0,0 +1,76 @@
...
+#include "antares-xpansion/benders/benders_core/BendersProblemFromFile.h"
+#include "antares-xpansion/benders/benders_core/ConstraintsFileReader.h"
+#include "antares-xpansion/benders/benders_core/SubproblemWorker.h"
+#include "antares-xpansion/benders/benders_core/Worker.h"
+#include "antares-xpansion/multisolver_interface/Solver.h"
+#include "antares-xpansion/xpansion_interfaces/ILogger.h"
+
+enum class CoeffType
+{
+    constraints = 0,
```
JasonMarechal25, line 19: niptick: Let's capitalize constants

data_test/low_memory_study/options.json
```diff
@@ -0,0 +1,29 @@
+{
+    "MASTER_NAME": "master",
+    "RELATIVE_GAP": 1.0e-6,
+    "MICRO_ITERATIONS": false,
+    "BOUND_ALPHA": true,
+    "MASTER_FORMULATION": "integer",
+    "LAST_MASTER_BASIS": "master_last_basis.bss",
+    "TRACE": true,
+    "STRUCTURE_FILE": "structure.txt",
+    "CACHE_PROBLEMS":2,
```
JasonMarechal25, line 10: issue: Instead of integer value we should opt for using literral that mach enum values. This would prove more clear when reading option file

JasonMarechal25, line 10: issue: Since "CACHE_PROBLEMS" was a boolean and we now have a third value, the name "CACHE_PROBLEMS" is not adequate anymore. We should:

- Keep "CACHE_PROBLEMS" as-is, a boolean value
- Deprecate "CACHE_PROBLEMS" by triggering a warning when a value is present
- Create a new entry for the optimization option
- Add a consistency check when both value are present

@tbittar unless we decide we remove CACHE_PROBLEMS entirely in favour of the new option?

src/cpp/benders/benders_core/include/antares-xpansion/benders/benders_core/BendersBase.h
```diff
@@ -161,7 +163,8 @@ class BendersBase
     VariableMap master_variable_map_;
     CouplingMap coupling_map_;
     VariableMap _problem_to_id;
-
+    std::vector<std::vector<std::string>> fixed_skeleton_subs_per_proc_;
+    std::shared_ptr<FixedSkeletonSubProblemBuilder> fixed_skeleton_subprob_builder_;
```
JasonMarechal25, line 167: nitpik: since it is instanciated in initializeProblem, you can probably use unique_ptr.

src/cpp/benders/benders_core/include/antares-xpansion/benders/benders_core/BendersBase.h
```diff
@@ -161,7 +163,8 @@ class BendersBase
     VariableMap master_variable_map_;
     CouplingMap coupling_map_;
     VariableMap _problem_to_id;
-
+    std::vector<std::vector<std::string>> fixed_skeleton_subs_per_proc_;
```
JasonMarechal25, line 166: suggestion: you may want to use type alias for the string, vector<string>, or vector<vector<string>> for clarity.

src/cpp/benders/benders_core/include/antares-xpansion/benders/benders_core/BendersBase.h
JasonMarechal25, line 167: issue: I only see an assignement in BendersMPI. Does it work for other kind of benders ?

src/cpp/benders/benders_core/include/antares-xpansion/benders/benders_core/FixedSkeletonSubProblemBuilder.h
```diff
@@ -0,0 +1,76 @@
...
+                                   std::shared_ptr<SolverAbstract> solver);
+
+    std::shared_ptr<SubproblemWorker> create_sub_solver_abstract(std::string sub_name,
+                                                                 VariableMap& variable_map,
+                                                                 double cut_coefficient_tolerance,
+                                                                 double slave_weight);
+    int get_sub_number();
+
+private:
+    Logger logger_;
```
JasonMarechal25, line 44: issue: you are creating a new instance of a logger here. At best you will copy an existing one in this value using constructor member initialization list, but you probably don't want a copy but a reference of any kind

src/cpp/benders/benders_core/include/antares-xpansion/benders/benders_core/FixedSkeletonSubProblemBuilder.h
```diff
@@ -0,0 +1,76 @@
...
+    std::map<std::string, std::vector<SolverRepresentedRows>> micro_iters_added_rows;
+    std::map<std::string, std::vector<double>> obj_coeffs_;
+    std::vector<int> rhs_row_indices_;
+    std::map<std::string, std::vector<double>> rhs_;
+    std::shared_ptr<SolverAbstract> solver_;
+    std::shared_ptr<BendersProblemFromFile> benders_problem_provider_;
+    std::vector<int> constraints_col_indices_;
+    std::vector<int> constraints_row_indices_;
+    std::vector<int> obj_col_indices_;
+    SolverLogManager solver_log_manager_;
```
JasonMarechal25, line 72: question: value, not ref ?

src/cpp/benders/benders_core/include/antares-xpansion/benders/benders_core/FixedSkeletonSubProblemBuilder.h
```diff
@@ -0,0 +1,76 @@
...
+    SolverIO solver_IO_;
```
JasonMarechal25, line 73: question: value, not ref?

src/cpp/benders/benders_mpi/BendersMPI.cpp
```diff
@@ -34,7 +36,30 @@ void BendersMpi::InitializeProblems()
 {
     MatchProblemToId();
     SubProblemNamesInCut subs_per_proc;
-    if (_options.CACHE_PROBLEMS)
+    switch (_options.CACHE_PROBLEMS)
+    {
+    case 2:
+    {
+        std::cout << "case fixed skeleton " << std::endl;
```
JasonMarechal25, line 43: Don't use std::cout

JasonMarechal25 (changes requested):

JasonMarechal25:
We need to update and add to the existing documentation

Ideally we shoul also write a design document.

data_test/low_memory_study/options.json
tbittar: Yes that is the plan, but we did not want to break the current behavior in production yet (I think this will need a whole PR to handle retrocompatibility), and I agree with the other remark that it should lead to another field name, something like `MEMORY_OPTIMIZATION_STRATEGY` or something more concise

tbittar: See also https://github.com/AntaresSimulatorTeam/antares-xpansion/pull/1250#discussion_r3386310329

src/cpp/benders/benders_core/include/antares-xpansion/benders/benders_core/BendersBase.h
tbittar: No... actually this is hard as this is at the boundary between R&D and production code, what strategy to use to be efficient in testing some stuff, and then not forget to be compatible with all combination of options we can have (and there are many of them...) ?

data_test/low_memory_study/options.json
ze-hedi: technically the memory optimization algo is a cache problem technique. Initially we had a seperate variable to handle it. But as the legacy CACHE_PROBLEM behaviour and the memory optimization behaviour are mutually exclusive, we decided to handle all these cases from one variable.
Also the input for a memory optimization study is different

tests/cpp/benders/MemOptimSubProblemBuilderTest.cpp
ze-hedi: the tests implicitly calls the methods when the class is instantiated. If we want to call these methods seperately, we can : 
- put these methods as public (not a proper design) 
- create a sub class that implement get methods for the data structures that read the csv to be able to do assert on them. (attributes should then be protected)

src/cpp/benders/benders_core/include/antares-xpansion/benders/benders_core/FixedSkeletonSubProblemBuilder.h
ze-hedi: i use the solver manager to instantiate a new solver. every new solver needs a seperate manger object ?

PR #1276, opened 2026-08-25
fix(agent): Update testing documentation

JasonMarechal25:
Fix documentation and enable finding documentation for claude

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -13,7 +13,17 @@ cmake --preset vcpkg
 cmake --build build
 ```
 
-Run tests: `cmake --preset vcpkg -DBUILD_TESTING=ON && cmake --build build && ctest --test-dir build`
+## Test
+
+```bash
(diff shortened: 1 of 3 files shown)


AGENTS.md
```diff
@@ -13,7 +13,17 @@ cmake --preset vcpkg
...
+## Test
+
+```bash
+cmake --preset vcpkg -DBUILD_TESTING=ON && cmake --build build
+ctest --test-dir build -L unit                    # unit tests (C++ and Python)
+ctest --test-dir build -L unit -E unit_launcher   # C++ unit tests only
+```
+
+Select tests by label, not by name pattern — C++ unit-test names are
+inconsistent, so `-R '^unit_'` silently skips about half of them.
```
tbittar, line 25: Already in `testing.md`

tbittar (approved):

PR #1277, opened 2026-08-25
Refactor: SubproblemWorkerFactory to integrate SubproblemSkeleton

JasonMarechal25:


src/cpp/benders/benders_by_batch/BendersByBatch.cpp
```diff
@@ -518,7 +518,9 @@ void BendersByBatch::GetCompactInMemCuts(SubProblemDataMap& subproblem_data_map,
     {
         auto name = kvp.first;
         auto variable_map = kvp.second;
-        double slave_weights = SubproblemWeight(_data.nsubproblem, name);
+        double slave_weights = SubproblemWeight(
```
ze-hedi: This change is actually wrong! the objective function of each subproblem should be scaled by the whole number of subprobme we are considered not the one on the proc

src/cpp/benders/benders_core/include/antares-xpansion/benders/benders_core/SkeletonCoefficientSet.h
```diff
@@ -24,7 +25,7 @@ class SkeletonCoefficientSet
               std::optional<std::filesystem::path> row_indices_csv,
               const std::shared_ptr<SolverAbstract>& solver,
               Logger& logger,
-              boost::mpi::communicator* world);
+              AbortFunc abort_func = nullptr);
```
flomnes, line 28: Keep in mind that calling the default instance for `std::function` raises `std::bad_function_call`

```cpp
#include <functional>

int main()
{
    std::function<int(void)> f;
    f(); // std::bad_function_call
}
```

Maybe you meant the no-op function ? `[] (int){}`

src/cpp/benders/benders_core/SubproblemSkeleton.cpp
```diff
@@ -0,0 +1,76 @@
...
+{
+    solver.chg_obj(all_col_indices_, Weighted(skeleton_obj_coeffs_, slave_weight));
+
+    solver.chg_coefs(coef_set_.RowIndices(),
+                     coef_set_.ColIndices(),
+                     coef_set_.CoefficientsFor(sub_name));
+    solver.chg_obj(obj_set_.ColIndices(),
+                   Weighted(obj_set_.CoefficientsFor(sub_name), slave_weight));
+    solver.chg_rhs_values(rhs_set_.RowIndices(), rhs_set_.CoefficientsFor(sub_name));
+}
```
flomnes, line 71: ### ⚠️  conflict ⚠️ 
See #1282

src/cpp/benders/plugins/Benders_MICRO_ITERS.cpp
```diff
@@ -582,12 +582,12 @@ void Benders_MICRO_ITERS::build_skeleton_constraint_set_loader(const BendersBase
...
-                                                                             constraints_names),
-                                                                           _world);
+    constraint_set_loader_ = std::make_shared<SkeletonConstraintSetLoader>(
+      options.INPUTROOT,
+      _logger,
+      options.SOLVER_NAME,
+      options.LOG_LEVEL,
+      options.PROBLEMS_FORMAT,
+      std::move(constraints_names),
+      _world ? AbortFunc([w = _world](int code) { w->abort(code); }) : AbortFunc{});
```
flomnes, line 592: No abort in sequential ? If confirmed, this gap is hard to justify

Suggestion : `std::exit(code)` when `_world == nullptr`

src/cpp/benders/benders_core/include/antares-xpansion/benders/benders_core/SkeletonCoefficientSet.h
```diff
@@ -1,18 +1,19 @@
...
 #include <string>
 #include <vector>
 
-#include <boost/mpi.hpp>
-
 #include "antares-xpansion/multisolver_interface/SolverAbstract.h"
 #include "antares-xpansion/xpansion_interfaces/ILogger.h"
 #include "skeleton_coefficient_reader.h"
 
+using AbortFunc = std::function<void(int)>;
```
YoshuaLisasi, line 15: maybe move this from here to its own file ( AbortFunc.h or in a namespace )  ? 
since it's used by many other classes such as SkeletonCoefficientSet, SkeletonConstraintSetLoader, SubproblemSkeleton, SubproblemWorkerFactory

src/cpp/benders/benders_core/include/antares-xpansion/benders/benders_core/SubproblemSkeleton.h
```diff
@@ -0,0 +1,51 @@
...
+ * loaded together and always applied together, which is why they live here
+ * rather than as loose members of SubproblemWorkerFactory.
+ */
+class SubproblemSkeleton
+{
+public:
+    /// Reads the six CSVs under `sub_dir`, keeping only the rows whose key is in
+    /// `sub_problem_names`. `solver` is used to resolve row/column names to indices
+    /// and to capture the skeleton's own objective, which ApplyTo() re-weights.
+    void Load(const std::filesystem::path& sub_dir,
```
YoshuaLisasi, line 27: This function has an implicit "unloaded" state , this class is default-constructable and has a Load method that is separated. 
This means that the behavior can be silently incorrect if ApplyTo is called before Load.
Maybe better to either have a constructor or add nad assert / bool_loaded.

src/cpp/benders/benders_core/SubproblemSkeleton.cpp
```diff
@@ -0,0 +1,76 @@
...
+
+#include <algorithm>
+#include <numeric>
+#include <utility>
+
+#include "antares-xpansion/benders/benders_core/skeleton_coefficient_reader.h"
+
+namespace
+{
+std::vector<double> Weighted(const std::vector<double>& coefficients, double slave_weight)
```
YoshuaLisasi, line 11: Maybe for an other PR : 

this function allocates a new vector each call.
In ApplyTo it's called 2 times , if the vectors are big it can weigh. 
An optimisation could be to have a reusable buffer as a member of SubproblemSkeleton instead of re-allocating everytime.
Not critical tho.

src/cpp/benders/benders_by_batch/BendersByBatch.cpp
YoshuaLisasi, line 521: slave_weights -> slave_weight ?

PR #1281, opened 2026-09-01
integrating best ub tracker in bendersmpi

ze-hedi:
feat: Track subproblem variable values at best upper bound

  Motivation

  In Benders decomposition, the best upper bound (best UB) iteration produces the best feasible
  solution found. Until now, there was no way to inspect the values of specific subproblem variables
  at that iteration — useful for post-processing, diagnostics, and warm-starting downstream tools.

  What this PR does

  Introduces BestUbTracker, a component integrated into the Benders MPI kernel that monitors a
  user-defined list of subproblem variables and records their values whenever a new best UB is found.

  Usage:
  - Provide a file sub_variables_to_save.csv in the study input root with a comma-separated list of
  variable names to track
  - At the end of the run, sub_best_ub_variables.csv is written in the output root with one row per
  subproblem and one column per tracked variable, reflecting their values at the last best UB
  iteration

  Implementation:
  - BestUbTracker reads the variable list at construction, lazily resolves variable indices per
  subproblem on first encounter, and updates stored solution vectors whenever a new best UB is found
  - At the end of the run, values are gathered across MPI ranks (rank 0 collects and writes the
  output)
  - Integrated into BendersMPI and BendersBase::GetCompactInMemCuts (compact memory / skeleton mode)

### DIFF
--- a/src/cpp/benders/benders_core/BestUbTracker.cpp
+++ b/src/cpp/benders/benders_core/BestUbTracker.cpp
@@ -1,0 +1,141 @@
+#include <antares-xpansion/benders/benders_core/BestUbTracker.h>
+
+#include <boost/tokenizer.hpp>
+
+#include "iostream"
+
+BestUbTracker::BestUbTracker(mpi::communicator* world,
+                                     const std::filesystem::path& file_path,
+                                     const std::filesystem::path& output_root,
+                                     Logger logger):
+    file_stream_(file_path),
+    _world(world),
+    _logger(logger)
+{
+    output_file_ = output_root / "sub_best_ub_variables.csv";
+    if (!file_stream_.is_open())
+    {
+        _logger->display_message("sub_variables_to_save.csv  not found ") ; 
+        return;
+
+    }
+
+    std::string line;
+    if (std::getline(file_stream_, line))
+    {
+        boost::escaped_list_separator<char> sep('\\', ',', '\"');
+        using Tokenizer = boost::tokenizer<boost::escaped_list_separator<char>>;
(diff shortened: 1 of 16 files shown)


src/cpp/benders/benders_core/BestUbTracker.cpp
```diff
@@ -0,0 +1,128 @@
...
+                if (index < 0)
+                {
+                    _logger->display_message("unable to find " + variable + " in sub_problem "
+                                             + sub_name);
+                }
+                variables_to_follow_indices_per_sub_[sub_name].push_back(index);
+            }
+        }
+
+        if (last_iteration_update_ == iter)
```
ze-hedi: this condition is obsolete as set_best_ub_solution_ and therefore the last_iteration_update_ class member  becomes obsolete

src/cpp/benders/benders_core/BestUbTracker.cpp
```diff
@@ -0,0 +1,128 @@
...
+        {
+            values_per_sub_[sub_name] = worker->get_solution();
```
ze-hedi: storing the whole solution is no good, subproblems can have a huge number of variables. it would be better if we trade some cpu compute and only get the values we are intrested in

src/cpp/benders/benders_core/BestUbTracker.cpp
```diff
@@ -0,0 +1,128 @@
...
+        }
+    }
+}
+
+void BestUbTracker::dump_values()
```
ze-hedi: after the changes related to the message just before the logic of this method will change

PR #1282, opened 2026-09-01
checking if csv files are empty or not before setting values on solve…

ze-hedi:
When building the building the solver abstract behind the subproblem worker we need to check if 
- rhs csv files are not empty 
- obj coeff files not empty 
- constraints csv files not empty

### DIFF
--- a/src/cpp/benders/benders_core/SubproblemWorkerFactory.cpp
+++ b/src/cpp/benders/benders_core/SubproblemWorkerFactory.cpp
@@ -116,30 +116,33 @@ std::shared_ptr<SubproblemWorker> SubproblemWorkerFactory::CreateSubSolverAbstra
 
     solver_->chg_obj(indices, weighted_obj);
 
-    /*we need to check if we are changing any constraints 
+    /*we need to check if we are changing any constraints
     It happens that the provided csv to reset constraints coeffs from sub to sub be empty
     If it's the case there's nothing to set on the solver to build the subproblem worker
     */
     if (!coef_set_.RowIndices().empty())
-        solver_->chg_coefs(coef_set_.RowIndices(),
-                       coef_set_.ColIndices(),
-                       coef_set_.CoefficientsFor(sub_name));
-
-    if (!obj_set_.CoefficientsFor(sub_name).empty()) 
     {
+        solver_->chg_coefs(coef_set_.RowIndices(),
+                           coef_set_.ColIndices(),
+                           coef_set_.CoefficientsFor(sub_name));
+    }
 
+    if (!obj_set_.CoefficientsFor(sub_name).empty())
+    {
         const auto& obj_coeffs = obj_set_.CoefficientsFor(sub_name);
         std::vector<double> weighted_obj_coeffs(obj_coeffs.size());
         std::ranges::transform(obj_coeffs,
-            weighted_obj_coeffs.begin(),
-            [slave_weights](double coefficient)
-            { return coefficient * slave_weights; });
-                
-            solver_->chg_obj(obj_set_.ColIndices(), weighted_obj_coeffs);
+                               weighted_obj_coeffs.begin(),
+                               [slave_weights](double coefficient)
+                               { return coefficient * slave_weights; });
+
+        solver_->chg_obj(obj_set_.ColIndices(), weighted_obj_coeffs);
     }
 
-    if (!rhs_set_.RowIndices().empty()) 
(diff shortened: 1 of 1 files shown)


tbittar (approved):

tbittar:
you could have added a simple unit test with empty RowIndices or CoefficientsFor to validate the behavior

src/cpp/benders/benders_core/SubproblemWorkerFactory.cpp
```diff
@@ -116,19 +116,33 @@ std::shared_ptr<SubproblemWorker> SubproblemWorkerFactory::CreateSubSolverAbstra
...
-    solver_->chg_rhs_values(rhs_set_.RowIndices(), rhs_set_.CoefficientsFor(sub_name));
+    /*we need to check if we are changing any constraints
+    It happens that the provided csv to reset constraints coeffs from sub to sub be empty
+    If it's the case there's nothing to set on the solver to build the subproblem worker
+    */
+    if (!coef_set_.RowIndices().empty())
+    {
+        solver_->chg_coefs(coef_set_.RowIndices(),
+                           coef_set_.ColIndices(),
+                           coef_set_.CoefficientsFor(sub_name));
```
flomnes, line 127: There is no consistency check on these 3 vectors. The `chg_coefs` loop only works if they have identical size
```cpp
void SolverCbc::chg_coefs(const std::vector<int>& id_rows,
                          const std::vector<int>& id_cols,
                          const std::vector<double>& vals)
{
    for (size_t i = 0; i < id_rows.size(); i++)
    {
        chg_coef(id_rows[i], id_cols[i], vals[i]);
    }
}
```
There could be cases where `coef_set_.RowIndices()` is not empty, but `coef_set_.ColIndices()` or `coef_set_.CoefficientsFor(sub_name)` is empty.

**Suggestion** add a check on vector sizes here or at reading time (`SkeletonCoefficientSet::Load`)

src/cpp/benders/benders_core/SubproblemWorkerFactory.cpp
```diff
@@ -116,19 +116,33 @@ std::shared_ptr<SubproblemWorker> SubproblemWorkerFactory::CreateSubSolverAbstra
...
+    }
+
+    if (!obj_set_.CoefficientsFor(sub_name).empty())
+    {
+        const auto& obj_coeffs = obj_set_.CoefficientsFor(sub_name);
+        std::vector<double> weighted_obj_coeffs(obj_coeffs.size());
```
flomnes, line 133: Same suggestion for `obj_set_.ColIndices()` and `obj_set_.CoefficientsFor(sub_name)`

Also you could perform one lookup instead of two here
```suggestion
    if (const auto& obj_coeffs = obj_set_.CoefficientsFor(sub_name); !obj_coeffs.empty())
    {
        
        std::vector<double> weighted_obj_coeffs(obj_coeffs.size());
```

src/cpp/benders/benders_core/SubproblemWorkerFactory.cpp
```diff
@@ -116,19 +116,33 @@ std::shared_ptr<SubproblemWorker> SubproblemWorkerFactory::CreateSubSolverAbstra
...
+                               [slave_weights](double coefficient)
+                               { return coefficient * slave_weights; });
+
+        solver_->chg_obj(obj_set_.ColIndices(), weighted_obj_coeffs);
+    }
+
+    if (!rhs_set_.RowIndices().empty())
+    {
+        solver_->chg_rhs_values(rhs_set_.RowIndices(), rhs_set_.CoefficientsFor(sub_name));
+    }
```
flomnes, line 145: No consistency check on the size of these two vectors, add one

Can perform one single lookup instead of 2 (assign `rhs_set_.RowIndices()` to a temporary)

flomnes (changes requested):
This probably fixes a real bug, but also reveals another one.

All three `chg_*` calls in `CreateSubSolverAbstract` invoke `CoefficientsFor(sub_name)` unconditionally without first checking whether the coefficient data for that subproblem exists. So when a coefficient CSV is empty/missing while the index CSVs are still populated, the lookup either aborts (MPI) or returns an empty vector, causing a size mismatch and out-of-bounds read in the subsequent `chg_*` call.

One style remark C++17 allows
```cpp
if (auto object = f(); object.method())
```
instead of
```cpp
auto object = f();
if (object.method())
```

PR #1285, opened 2026-09-04
Robust Resume mode

ze-hedi:
# feat(benders): robust resume mode — separate RESUME and HOT_START semantics

## Summary

- **Replaces the boolean `RESUME` flag with a three-value `ResumeMode` enum** (`COLD_START`, `RESUME`, `HOT_START`), giving explicit semantics to each re-entry scenario in the Benders algorithm.
- **`RESUME` mode** continues a partially completed simulation from an intermediate iteration — it appends to CSV logs, loads previous JSON output state (best UB, iteration data), and reads the saved master basis/cuts.
- **`HOT_START` mode** starts a fresh simulation (reinitialised output files, iteration count reset to 0) but reads an enriched `master.mps` that already contains accumulated cuts from a prior run, enabling convergence in far fewer iterations.
- **New Cucumber end-to-end tests** validate both modes: resume verifies continued convergence across stop/restart, and hot start verifies that a fully-converged master reconverges in exactly 1 iteration.

## Key changes

| Area | What changed |
|---|---|
| **Core type** | New `ResumeMode` enum in `ResumeMode.h` with string/int parsing (`resumeModeFromString`) |
| **Options** | `SimulationOptions` and `BendersBaseOptions` changed from `bool RESUME` to `ResumeMode RESUME` with a template specialization for deserialization |
| **C++ logic** | `BendersBase`, `StartUp`, `WriterFactories`, `BendersMPI`, `BendersSequential` — branching on enum values instead of bool; HOT_START shares master-read path with RESUME but does **not** load previous output state |
| **Python** | `resume_study.py` now writes `"resume"` (string) instead of `true` (bool) to the options JSON |
| **Test data** | 4 JSON option files updated from `"RESUME": true` to `"RESUME": "resume"` |
| **E2E tests** | New `benders_resume_modes.feature` with Resume and Hot Start scenarios + supporting step definitions |
| **Unit tests** | All C++ tests updated to use `ResumeMode::COLD_START` / `ResumeMode::RESUME` |

### DIFF
--- a/src/cpp/benders/benders_mpi/BendersMPI.cpp
+++ b/src/cpp/benders/benders_mpi/BendersMPI.cpp
@@ -156,7 +156,7 @@ void BendersMpi::InitializeMaster()
                                    get_log_level(),
                                    _data.nsubproblem,
                                    solver_log_manager_,
-                                   IsResumeMode(),
+                                   _options.RESUME != 0,
                                    _logger,
                                    Options().PROBLEMS_FORMAT,
                                    benders_problem_provider.get(),
--- a/src/cpp/benders/benders_mpi/BendersMPI.cpp
+++ b/src/cpp/benders/benders_mpi/BendersMPI.cpp
@@ -156,7 +156,7 @@ void BendersMpi::InitializeMaster()
                                    get_log_level(),
                                    _data.nsubproblem,
                                    solver_log_manager_,
-                                   _options.RESUME != 0,
+                                   _options.RESUME != ResumeMode::COLD_START,
                                    _logger,
                                    Options().PROBLEMS_FORMAT,
                                    benders_problem_provider.get(),
--- a/src/cpp/benders/benders_sequential/BendersSequential.cpp
+++ b/src/cpp/benders/benders_sequential/BendersSequential.cpp
@@ -40,7 +40,7 @@ void BendersSequential::InitializeProblems()
                                get_log_level(),
                                _data.nsubproblem,
                                solver_log_manager_,
-                               _options.RESUME != 0,
+                               _options.RESUME != ResumeMode::COLD_START,
                                _logger,
                                Options().PROBLEMS_FORMAT,
                                benders_problem_provider.get(),
--- a/src/cpp/benders/benders_sequential/BendersSequential.cpp
+++ b/src/cpp/benders/benders_sequential/BendersSequential.cpp
@@ -40,7 +40,7 @@ void BendersSequential::InitializeProblems()
                                get_log_level(),
                                _data.nsubproblem,
                                solver_log_manager_,
-                               IsResumeMode(),
+                               _options.RESUME != 0,
                                _logger,
                                Options().PROBLEMS_FORMAT,
                                benders_problem_provider.get(),
--- a/src/cpp/benders/factories/include/antares-xpansion/benders/factories/WriterFactories.h
+++ b/src/cpp/benders/factories/include/antares-xpansion/benders/factories/WriterFactories.h
@@ -10,6 +10,6 @@
 std::shared_ptr<Output::OutputWriter> build_void_writer();
 
 std::shared_ptr<Output::OutputWriter> build_json_writer(const std::filesystem::path& json_file_name,
-                                                        bool restart);
+                                                        const int restart);
 
 #endif // ANTARESXPANSION_WRITERFACTORIES_H
--- a/src/cpp/benders/factories/include/antares-xpansion/benders/factories/WriterFactories.h
+++ b/src/cpp/benders/factories/include/antares-xpansion/benders/factories/WriterFactories.h
@@ -5,11 +5,12 @@
 #include <filesystem>
 #include <string>
 
+#include "antares-xpansion/benders/benders_core/common.h"
 #include "antares-xpansion/benders/output/OutputWriter.h"
 
 std::shared_ptr<Output::OutputWriter> build_void_writer();
 
 std::shared_ptr<Output::OutputWriter> build_json_writer(const std::filesystem::path& json_file_name,
-                                                        const int restart);
+                                                        ResumeMode restart);
 
 #endif // ANTARESXPANSION_WRITERFACTORIES_H
--- a/src/cpp/core/include/antares-xpansion/core/ResumeMode.h
+++ b/src/cpp/core/include/antares-xpansion/core/ResumeMode.h
@@ -1,0 +1,47 @@
+#pragma once
+
+#include <ostream>
+#include <stdexcept>
+#include <string>
+
+enum class ResumeMode
+{
+    COLD_START = 0,
+    RESUME = 1,
+    HOT_START = 2
+};
+
+inline ResumeMode resumeModeFromString(const std::string& str)
+{
+    if (str == "0" || str == "cold_start")
+    {
+        return ResumeMode::COLD_START;
+    }
+    if (str == "1" || str == "resume")
+    {
+        return ResumeMode::RESUME;
+    }
+    if (str == "2" || str == "hot_start")
+    {
+        return ResumeMode::HOT_START;
+    }
+    throw std::runtime_error("Unknown ResumeMode: " + str
+                             + " (expected 0, 1, 2, cold_start, resume, or hot_start)");
+}
+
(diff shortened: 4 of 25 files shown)


ze-hedi (commented):
border case to consider : when we are on hot start case, and that we relaunch the same study (we add nothing in the master or change subproblem) the user will maybe waiting  have to have one iteration. 
However as we don't have the info on the bestub (that can be different from the last iteration at the first run) we risk having some iterations (until we find back the bestub since the codition of stop is bestub-lb) 
As in the hot start, with the current implementation we completly ignore the out.json of the precedent run so we don't have the info about the last bestub. 
The fix is simple: check if there's an out.json even in hot_strart mode, if yes we go and look for the bestub (may be only if the problem_status is OPTIMAL which means we already converged at the precedent run and therefore the stored best ub is the "real one")

src/cpp/core/include/antares-xpansion/core/ResumeMode.h
```diff
@@ -0,0 +1,47 @@
...
+        return ResumeMode::COLD_START;
+    }
+    if (str == "1" || str == "resume")
+    {
+        return ResumeMode::RESUME;
+    }
+    if (str == "2" || str == "hot_start")
+    {
+        return ResumeMode::HOT_START;
+    }
```
tbittar: Do we need to keep compatibility with 0,1, 2 ? @JasonMarechal25 @ze-hedi 
I advocate to only allow resume, hot start, and cold start to simplify

tbittar (changes requested):
You need to update the Benders options documentation

JasonMarechal25:
Are resume and hot/cold start really the same kind of feature ?
Resume was meant to continue a stopped simulation. Hot start feels more like an optimisation option.

tbittar:
Maybe ResumeMode is not the correct option name but we want to handle several use cases with this option:
- Cold start : Classical Benders use case starting from scratch
- Resume : A previous simulation was interrrupted for some reason, you want to continue the run starting from master_last_iteration you already have + all log files you will complete (append mode)
- Hot start : A previous simulation has been successfully completed. But then, by an external process, you modify slightly your master_last_iteration to add / modify a few constraints. You want to benefit from all cuts already present from the previous run to speed up the Benders resolution. Basically, from Benders point of view this is exactly as the resume mode, but this time you want clean logs to start with.

tbittar:
Indeed we have several notions of hot start :
- "Benders-level" hot start, that is starting the whole Benders algorithm with a master that already has some cuts (tackled in this PR)
- "Subproblem-level" hot start, that is hot starting subproblem resolution between two iterations using the previous basis

src/cpp/core/include/antares-xpansion/core/ResumeMode.h
```diff
@@ -0,0 +1,47 @@
...
+#include <string>
+
+enum class ResumeMode
+{
+    COLD_START = 0,
+    RESUME = 1,
+    HOT_START = 2
+};
+
+inline ResumeMode resumeModeFromString(const std::string& str)
```
YoshuaLisasi: The description says  
" New ResumeMode enum in ResumeMode.h with string/int parsing (resumeModeFromString) " 

But only the string part is define in the code.

src/cpp/core/include/antares-xpansion/core/ResumeMode.h
YoshuaLisasi: **Question: master basis handling in HOT_START mode**

With this change, `HOT_START` goes through the same master-read path as `RESUME`. Could you confirm what this flag triggers inside the master initialization? In particular, does it also load the saved master basis (`LAST_MASTER_BASIS`)?

If it does, I think there's an edge case worth handling. In `RESUME`, the master and the basis are always consistent, because they were written at the same iteration of the same run. In `HOT_START`, we start a new simulation from a provided `master.mps`, and nothing guarantees that the basis file on disk matches it:

- **Missing basis:** the enriched master comes from another study, the output folder was cleaned, or the previous run stopped before writing the basis.
- **Mismatched basis:** e.g. run A converges and we keep its master; then run B (cold start) overwrites `master_last_basis`; a hot start with A's master would then load B's basis, which has different cuts/rows.

Since the basis is only a warm-start point, this shouldn't affect the optimal solution. However, depending on the solver it could raise an error, or behave differently across solvers (Xpress vs Coin), and that behavior isn't specified.

The current E2E test only covers the happy path: the master and the basis both come from the same converged run, so this case isn't exercised.

src/cpp/benders/benders_mpi/BendersMPI.cpp
```diff
@@ -156,7 +156,7 @@ void BendersMpi::InitializeMaster()
                                    get_log_level(),
                                    _data.nsubproblem,
                                    solver_log_manager_,
-                                   IsResumeMode(),
+                                   _options.RESUME != ResumeMode::COLD_START,
```
YoshuaLisasi: This code line is duplicated.
It's not clear at first glance why " everything except cold start".
Maybe try adding these functions : 


```
bool BendersBase::IsResumeMode() const
{
    return _options.RESUME == ResumeMode::RESUME;
}

bool BendersBase::ShouldReadPreviousMaster() const
{
    return _options.RESUME != ResumeMode::COLD_START;
}
```

src/cpp/benders/benders_sequential/BendersSequential.cpp
```diff
@@ -40,7 +40,7 @@ void BendersSequential::InitializeProblems()
                                get_log_level(),
                                _data.nsubproblem,
                                solver_log_manager_,
-                               IsResumeMode(),
+                               _options.RESUME != ResumeMode::COLD_START,
```
YoshuaLisasi: Same comment

src/cpp/benders/factories/include/antares-xpansion/benders/factories/WriterFactories.h
```diff
@@ -5,11 +5,12 @@
...
 #include <string>
 
+#include "antares-xpansion/benders/benders_core/common.h"
 #include "antares-xpansion/benders/output/OutputWriter.h"
 
 std::shared_ptr<Output::OutputWriter> build_void_writer();
 
 std::shared_ptr<Output::OutputWriter> build_json_writer(const std::filesystem::path& json_file_name,
-                                                        bool restart);
+                                                        ResumeMode restart);
```
YoshuaLisasi: Maybe rename restart to resume_mode

src/cpp/benders/factories/include/antares-xpansion/benders/factories/WriterFactories.h
```diff
@@ -5,11 +5,12 @@
 #include <filesystem>
 #include <string>
 
+#include "antares-xpansion/benders/benders_core/common.h"
```
YoshuaLisasi: benders_core/common.h is included only for the enum 
Maybe try to add only : antares-xpansion/core/ResumeMode.h

src/cpp/benders/benders_mpi/BendersMPI.cpp
ze-hedi: i think it would be better if we just delete the IsResumeMode and replace it with the if condition directly.

src/cpp/core/include/antares-xpansion/core/ResumeMode.h
ze-hedi: In the case of the HOT_START, the master problem can contain more constrain which lead to changing the whole basis so setting the basis of the previous run can be problematic.
@tbittar do you confirm that ?

tbittar: Ah good remark !
- If the basis is missing, we need to ensure that the code does not crash, but then we just solve the master without hot start and that should be fine
- In HOT_START mode, I think the idea is indeed to add constraints into the master compared to the previous run, so I think it is safer not to load the master basis, and the solver may take time to make it feasible. It will not change the optimal solution, but may impact the solution (loading the master basis could be beneficial or not depending on the use case, this is impossible to predict a priori, so I advocate not to load it). More refined work would be possible to deduce a basis given the previous and the master modification : out of scope here (needs R&D work)
- In RESUME mode, loading the master basis is always beneficial (as we have exactly the same master as previously)

PR #1286, opened 2026-09-04
refactor: Replace int CACHE_PROBLEMS with CacheProblems enum class

ze-hedi:
## Summary

This PR replaces the raw `int` type used for the `CACHE_PROBLEMS` option with a proper `enum class CacheProblems { NO_CACHE, PER_SUB, COMPACT }`. The primary motivation is to enhance code readability and maintainability: scattered integer literals (`0`, `1`, `2`) and comparisons like `> 0` or `< 2` throughout the benders codebase made it difficult to understand the intent of each branch at a glance. With named enum values, every switch case and conditional now clearly expresses which caching strategy it targets, making the code self-documenting and reducing the risk of errors when extending or reviewing the logic.

## Changes

- Add `CacheProblems.h` defining the enum class with a case-insensitive `cacheProblemsFromString()` converter (supports both named values and legacy integer strings for backward compatibility) and `operator<<` for stream output
- Wire the new type into the `BENDERS_OPTIONS_MACRO` system (`SimulationOptions.hxx`) and add `Json::Value::as<CacheProblems>()` template specialization (`SimulationOptions.cpp`)
- Update `BendersBaseOptions::CACHE_PROBLEMS` field in `common.h` from `int` to `CacheProblems`
- Replace all integer literals and comparisons in `BendersBase.cpp`, `BendersByBatch.cpp`, `BendersMPI.cpp`, and `Benders_MICRO_ITERS.cpp`

## Test plan

- [x] Verify the project compiles successfully
- [x] Run existing end-to-end tests with `CACHE_PROBLEMS` set to `0`, `1`, and `2` (integer values in options.json) to confirm backward compatibility
- [x] Run existing end-to-end tests with `CACHE_PROBLEMS` set to `"NO_CACHE"`, `"PER_SUB"`, `"COMPACT"` (string values) to confirm new format works

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/cpp/core/include/antares-xpansion/core/CacheProblems.h
+++ b/src/cpp/core/include/antares-xpansion/core/CacheProblems.h
@@ -1,0 +1,52 @@
+#pragma once
+
+#include <algorithm>
+#include <ostream>
+#include <stdexcept>
+#include <string>
+
+#include "antares-xpansion/xpansion_interfaces/StringManip.h"
+
+enum class CacheProblems
+{
+    NO_CACHE,
+    PER_SUB,
+    COMPACT
+};
+
+inline CacheProblems cacheProblemsFromString(const std::string& str)
+{
+    auto lower_str = StringManip::StringUtils::ToLowercase(str);
+    if (lower_str == "no_cache" || lower_str == "0")
+    {
+        return CacheProblems::NO_CACHE;
+    }
+    if (lower_str == "per_sub" || lower_str == "1")
+    {
+        return CacheProblems::PER_SUB;
+    }
+    if (lower_str == "compact" || lower_str == "2")
+    {
+        return CacheProblems::COMPACT;
+    }
+    throw std::runtime_error("Unknown CacheProblems: " + str);
+}
+
+inline std::ostream& operator<<(std::ostream& stream, const CacheProblems& rhs)
+{
+    switch (rhs)
+    {
+    case CacheProblems::NO_CACHE:
+        stream << "NO_CACHE";
+        break;
+    case CacheProblems::PER_SUB:
+        stream << "PER_SUB";
+        break;
+    case CacheProblems::COMPACT:
+        stream << "COMPACT";
+        break;
+    default:
+        stream << "Unknown";
+    }
+    return stream;
+}
--- a/src/python/antares_xpansion/input_parser.py
+++ b/src/python/antares_xpansion/input_parser.py
@@ -92,9 +92,9 @@ class InputParser:
                                  help="Runs presolve step before Benders methods (Xpress only)")
         self.parser.add_argument("--cache_problems",
                                  dest=LauncherOptionsKeys.cache_problems_key(),
-                                 default=False,
-                                 action='store_true',
-                                 help="Cache problems on disque during benders")
+                                 default="NO_CACHE",
+                                 choices=["NO_CACHE", "PER_SUB", "COMPACT"],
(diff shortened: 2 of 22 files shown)


src/python/antares_xpansion/input_parser.py
```diff
@@ -92,9 +92,9 @@ def _initialize_parser(self):
                                  help="Runs presolve step before Benders methods (Xpress only)")
         self.parser.add_argument("--cache_problems",
                                  dest=LauncherOptionsKeys.cache_problems_key(),
-                                 default=False,
-                                 action='store_true',
-                                 help="Cache problems on disque during benders")
+                                 default="NO_CACHE",
+                                 choices=["NO_CACHE", "PER_SUB", "COMPACT"],
+                                 help="Cache problems mode during benders (NO_CACHE, PER_SUB, COMPACT)")
         self.parser.add_argument("--problem-format",
```
tbittar, line 98: @JasonMarechal25 This means we should update the command line in the launcher and synchronise deployment ?

tbittar (approved):

src/cpp/core/include/antares-xpansion/core/CacheProblems.h
```diff
@@ -0,0 +1,52 @@
...
+        stream << "PER_SUB";
+        break;
+    case CacheProblems::COMPACT:
+        stream << "COMPACT";
+        break;
+    default:
+        stream << "Unknown";
+    }
+    return stream;
+}
```
flomnes, line 52: If this is only used in tests, you can move this function to the test source file

If it is not used, remove it (avoid dead code)

src/cpp/core/include/antares-xpansion/core/CacheProblems.h
```diff
@@ -0,0 +1,52 @@
...
+        return CacheProblems::NO_CACHE;
+    }
+    if (lower_str == "per_sub" || lower_str == "1")
+    {
+        return CacheProblems::PER_SUB;
+    }
+    if (lower_str == "compact" || lower_str == "2")
+    {
+        return CacheProblems::COMPACT;
+    }
```
flomnes: Accepting legacy values here is misleading, I'd rather migrate existing input data

```suggestion
    if (lower_str == "no_cache")
    {
        return CacheProblems::NO_CACHE;
    }
    if (lower_str == "per_sub")
    {
        return CacheProblems::PER_SUB;
    }
    if (lower_str == "compact")
    {
        return CacheProblems::COMPACT;
    }
```

flomnes (commented):
src/python/antares_xpansion/trajectory/driver_resolution.py still has
```py
    cache_problems: bool
```
consider adjusting the type annotation to `str` to match the possible choices

src/cpp/core/include/antares-xpansion/core/CacheProblems.h
ze-hedi: we nee it in the print in simulationOptions.h

YoshuaLisasi (approved):

JasonMarechal25 (approved):

src/python/antares_xpansion/input_parser.py
JasonMarechal25: yes

JasonMarechal25: https://github.com/orgs/AntaresSimulatorTeam/projects/22/views/1?query=sort%3Aupdated-desc+state%3Aopen

PR #1287, opened 2026-09-05
Replacing micro_iterations_config.txt by micro_iterations_config.json

ze-hedi:
So far, the parameters of the micro iterations mode were handeled by reading a txt file.
However such format is harder to read and to get right so we changed it with a json file

### DIFF
--- a/data_test/ieee96_micro_it/micro_iterations_config.json
+++ b/data_test/ieee96_micro_it/micro_iterations_config.json
@@ -1,0 +1,4 @@
+{
+    "plugin_lib_path":"plugin_inputs/dummy_micro_iterations_plugin.so" , 
+    "warm_start":0
+}
--- a/data_test/ieee96_micro_it_skeleton_study/micro_iterations_config.json
+++ b/data_test/ieee96_micro_it_skeleton_study/micro_iterations_config.json
@@ -1,0 +1,4 @@
+{
+    "plugin_lib_path":"plugin_inputs/dummy_micro_iterations_plugin.so" , 
+    "warm_start":0
+}
--- a/src/cpp/benders/plugins/Benders_MICRO_ITERS.cpp
+++ b/src/cpp/benders/plugins/Benders_MICRO_ITERS.cpp
@@ -13,6 +13,7 @@
 #include <string_view>
 
 #include <boost/tokenizer.hpp>
+#include <json/reader.h>
 
 #include "iostream"
 
@@ -141,37 +142,32 @@ void Benders_MICRO_ITERS::read_micro_iteration_config_file()
 {
     // Reading the micro iterations configuration file
     std::filesystem::path mirco_iterations_options_path = std::filesystem::path(options_.INPUTROOT)
-                                                          / "micro_iterations_config.txt";
+                                                          / "micro_iterations_config.json";
     std::ifstream micro_iterations_options_stream(mirco_iterations_options_path.string());
 
     if (micro_iterations_options_stream.is_open())
     {
-        std::string line;
-        while (std::getline(micro_iterations_options_stream, line))
+        Json::Value config;
+        Json::CharReaderBuilder reader_builder;
+        std::string errors;
+        if (!Json::parseFromStream(reader_builder, micro_iterations_options_stream, &config, &errors))
         {
-            std::istringstream iss(line);
-            std::string key, value;
+            std::cerr << "Failed to parse JSON config: " << errors << std::endl;
+            exit(EXIT_FAILURE);
+        }
 
-            if (std::getline(iss, key, '=') && std::getline(iss, value))
+        for (const auto& key : config.getMemberNames())
+        {
+            if (key == "warm_start")
             {
-                if (key == "warm_start")
+                if (config[key].asString() == "0")
                 {
-                    if (value == "0")
-                    {
-                        warm_start_ = false;
-                    }
-                }
-                else
-                {
-                    micro_iterations_config_[key] = value;
+                    warm_start_ = false;
                 }
             }
-        }
-        for (auto [key, value]: micro_iterations_config_)
-        {
-            if (key == "plugin_lib_path")
+            else
             {
-                std::string cpp_output_lib_path = micro_iterations_config_["plugin_lib_path"];
+                micro_iterations_config_[key] = config[key].asString();
             }
         }
     }
--- a/src/cpp/benders/plugins/Benders_MICRO_ITERS.cpp
+++ b/src/cpp/benders/plugins/Benders_MICRO_ITERS.cpp
@@ -8,12 +8,12 @@
 #include <chrono>
 #include <exception>
 #include <fstream>
+#include <json/reader.h>
 #include <sstream>
 #include <stdexcept>
 #include <string_view>
 
 #include <boost/tokenizer.hpp>
-#include <json/reader.h>
 
 #include "iostream"
 
@@ -150,13 +150,16 @@ void Benders_MICRO_ITERS::read_micro_iteration_config_file()
         Json::Value config;
         Json::CharReaderBuilder reader_builder;
         std::string errors;
-        if (!Json::parseFromStream(reader_builder, micro_iterations_options_stream, &config, &errors))
+        if (!Json::parseFromStream(reader_builder,
+                                   micro_iterations_options_stream,
+                                   &config,
+                                   &errors))
         {
             std::cerr << "Failed to parse JSON config: " << errors << std::endl;
             exit(EXIT_FAILURE);
         }
 
-        for (const auto& key : config.getMemberNames())
+        for (const auto& key: config.getMemberNames())
         {
             if (key == "warm_start")
             {
--- a/src/cpp/benders/plugins/CMakeLists.txt
+++ b/src/cpp/benders/plugins/CMakeLists.txt
@@ -9,9 +9,10 @@ target_sources(plugins PRIVATE
 )
 
 target_link_libraries(plugins
-            PUBLIC 
+            PUBLIC
             antaresXpansion::benders_core
             antaresXpansion::logger_lib
+            ${JSONCPP_LIB}
 )
 
 target_include_directories(plugins
--- a/tests/end_to_end/cucumber/features/steps/given.py
+++ b/tests/end_to_end/cucumber/features/steps/given.py
@@ -53,14 +53,11 @@ def set_cache_problems_level(context, level):
 
 @given("the warm start is {value}")
 def set_warm_start(context, value):
-    config_path = context.tmp_study / "micro_iterations_config.txt"
-    match = False
-    with fileinput.FileInput(str(config_path), inplace=True) as file:
-        for line in file:
-            match = match or re.search(r"warm_start\s*=.*", line)
-            print(re.sub(r"warm_start\s*=.*", f"warm_start={value}", line), end="")
-    if not match:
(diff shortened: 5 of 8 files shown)


src/cpp/benders/plugins/Benders_MICRO_ITERS.cpp
```diff
@@ -141,37 +142,35 @@ void Benders_MICRO_ITERS::read_micro_iteration_config_file()
...
+            exit(EXIT_FAILURE);
+        }
 
-            if (std::getline(iss, key, '=') && std::getline(iss, value))
+        for (const auto& key: config.getMemberNames())
+        {
+            if (key == "warm_start")
             {
-                if (key == "warm_start")
+                if (config[key].asString() == "0")
```
ze-hedi: replace this with a read a bolean since warm start is kind of true or false

src/cpp/benders/plugins/Benders_MICRO_ITERS.cpp
```diff
@@ -141,37 +142,35 @@ void Benders_MICRO_ITERS::read_micro_iteration_config_file()
...
+        std::string errors;
+        if (!Json::parseFromStream(reader_builder,
+                                   micro_iterations_options_stream,
+                                   &config,
+                                   &errors))
         {
-            std::istringstream iss(line);
-            std::string key, value;
+            std::cerr << "Failed to parse JSON config: " << errors << std::endl;
+            exit(EXIT_FAILURE);
```
flomnes: 1. Print the full path to the JSON file here
2. Use a logger if possible instead of `std::cerr`

src/cpp/benders/plugins/Benders_MICRO_ITERS.cpp
flomnes: It would be better to use bool in the JSON, so you just call JSON's bool `asBool` (~~provided it exists~~ https://open-source-parsers.github.io/jsoncpp/class_json_1_1_value.html#ab693fb7b9b1595bb0adc49658bbf780d)
```json
{
    "warm_start": false
}
```

src/cpp/benders/plugins/Benders_MICRO_ITERS.cpp
```diff
@@ -141,37 +142,35 @@ void Benders_MICRO_ITERS::read_micro_iteration_config_file()
...
+                    warm_start_ = false;
                 }
             }
-        }
-        for (auto [key, value]: micro_iterations_config_)
-        {
-            if (key == "plugin_lib_path")
+            else
             {
-                std::string cpp_output_lib_path = micro_iterations_config_["plugin_lib_path"];
```
flomnes, line 174: The new code is fine, but I don't understand how `cpp_output_lib_path` was used, it seems to be a no-op
1. Get a value from a map
2. Store it to a temporary string
3. The string goes out of scope

src/cpp/benders/plugins/Benders_MICRO_ITERS.cpp
flomnes: typo here
```suggestion
        std::cerr << "unable to open : " << micro_iterations_options_path.string() << std::endl;
```

data_test/ieee96_micro_it/micro_iterations_config.json
flomnes, line 1: formatting

data_test/ieee96_micro_it_skeleton_study/micro_iterations_config.json
flomnes, line 1: formatting

flomnes (changes requested):

data_test/ieee96_micro_it/micro_iterations_config.json
tbittar, line 1: Why not reusing the symbolic link trick to avoid duplicating the json file ?

src/cpp/benders/plugins/Benders_MICRO_ITERS.cpp
tbittar: Also use a logger

tests/end_to_end/cucumber/features/steps/given.py
```diff
@@ -53,14 +53,11 @@ def set_cache_problems_level(context, level):
...
-        for line in file:
-            match = match or re.search(r"warm_start\s*=.*", line)
-            print(re.sub(r"warm_start\s*=.*", f"warm_start={value}", line), end="")
-    if not match:
-        with open(config_path, "a") as file:
-            file.write(f"\nwarm_start={value}")
+    config_path = context.tmp_study / "micro_iterations_config.json"
+    with open(str(config_path), "r") as file:
+        config = json.load(file)
+    config["warm_start"] = int(value)
```
tbittar, line 59: Set a boolean

tbittar: This may impact your plugin implementation ?

tbittar (changes requested):

data_test/ieee96_micro_it_skeleton_study/micro_iterations_config.json
ze-hedi: what's the issue with the formatting

src/cpp/benders/plugins/Benders_MICRO_ITERS.cpp
```diff
@@ -141,43 +142,39 @@ void Benders_MICRO_ITERS::read_micro_iteration_config_file()
...
+        std::string errors;
+        if (!Json::parseFromStream(reader_builder,
+                                   micro_iterations_options_stream,
+                                   &config,
+                                   &errors))
         {
-            std::istringstream iss(line);
-            std::string key, value;
+            _logger->display_message("failed to open : " + mirco_iterations_options_path.string());
+            _logger->display_message("Erros :" + errors);
```
flomnes: 1. Same log (one single `_logger->display_message`)
2. Erros -> Errors

tests/end_to_end/cucumber/features/steps/given.py
ze-hedi: no these values are only red on the benders side

data_test/ieee96_micro_it/micro_iterations_config.json
YoshuaLisasi: Maybe change the doc to tell user about these changes

src/cpp/benders/plugins/Benders_MICRO_ITERS.cpp
```diff
@@ -141,43 +142,41 @@ void Benders_MICRO_ITERS::read_micro_iteration_config_file()
...
-            std::string key, value;
+            std::ostringstream oss;
+            oss << "failed to open : " << mirco_iterations_options_path.string() << "\n"
+                << "Errors: " << errors;
+            _logger->display_message(oss.str());
+            exit(EXIT_FAILURE);
+        }
 
-            if (std::getline(iss, key, '=') && std::getline(iss, value))
+        for (const auto& key: config.getMemberNames())
```
YoshuaLisasi: Maybe check 

`if (!config.isObject())`

src/cpp/benders/plugins/Benders_MICRO_ITERS.cpp
```diff
@@ -141,43 +142,41 @@ void Benders_MICRO_ITERS::read_micro_iteration_config_file()
...
+        std::string errors;
+        if (!Json::parseFromStream(reader_builder,
+                                   micro_iterations_options_stream,
+                                   &config,
+                                   &errors))
         {
-            std::istringstream iss(line);
-            std::string key, value;
+            std::ostringstream oss;
+            oss << "failed to open : " << mirco_iterations_options_path.string() << "\n"
```
YoshuaLisasi: More failed to parse than faied to open ?

src/cpp/benders/plugins/Benders_MICRO_ITERS.cpp
```diff
@@ -141,43 +142,41 @@ void Benders_MICRO_ITERS::read_micro_iteration_config_file()
...
-                    if (value == "0")
-                    {
-                        warm_start_ = false;
-                    }
-                }
-                else
-                {
-                    micro_iterations_config_[key] = value;
-                }
+                warm_start_ = config[key].asBool();
```
YoshuaLisasi: Maybe test isBool() before and have an error message

src/cpp/benders/plugins/Benders_MICRO_ITERS.cpp
```diff
@@ -141,43 +142,41 @@ void Benders_MICRO_ITERS::read_micro_iteration_config_file()
...
             }
-        }
-        for (auto [key, value]: micro_iterations_config_)
-        {
-            if (key == "plugin_lib_path")
+            else
             {
-                std::string cpp_output_lib_path = micro_iterations_config_["plugin_lib_path"];
+                micro_iterations_config_[key] = config[key].asString();
```
YoshuaLisasi: Maybe test isString() before and have an error message

src/cpp/benders/plugins/Benders_MICRO_ITERS.cpp
```diff
@@ -141,43 +142,41 @@ void Benders_MICRO_ITERS::read_micro_iteration_config_file()
...
+            oss << "failed to open : " << mirco_iterations_options_path.string() << "\n"
+                << "Errors: " << errors;
+            _logger->display_message(oss.str());
+            exit(EXIT_FAILURE);
+        }
 
-            if (std::getline(iss, key, '=') && std::getline(iss, value))
+        for (const auto& key: config.getMemberNames())
+        {
+            if (key == "warm_start")
```
YoshuaLisasi: - Maybe have a warning if there are keys other than the ones we are expecting.
- And add an error if a require key is not there : 
```
 if (!config.isMember("plugin_lib_path"))
{
    _logger->display_message(path.string() + " : missing required key 'plugin_lib_path'");
    exit(EXIT_FAILURE);
}
```

src/cpp/benders/plugins/CMakeLists.txt
```diff
@@ -9,9 +9,10 @@ target_sources(plugins PRIVATE
 )
 
 target_link_libraries(plugins
-            PUBLIC 
+            PUBLIC
             antaresXpansion::benders_core
             antaresXpansion::logger_lib
+            ${JSONCPP_LIB}
```
YoshuaLisasi: Question : why not ? 

```
target_link_libraries(plugins
    PUBLIC
        antaresXpansion::benders_core
        antaresXpansion::logger_lib
    PRIVATE
        ${JSONCPP_LIB}
)
```

ze-hedi: great point ! the writer is only used in .cpp no risk of links to consumers of plugins

src/cpp/benders/plugins/Benders_MICRO_ITERS.cpp
ze-hedi: for the warm start i think it would be better if we launch the simulation but with warm start = false. for the plugin_lib_path if it's not given the handle (line 44) will be null and we will throw an error. but i agree that it would be better to check if it's empty or not before even doing the dlopen

src/cpp/benders/plugins/Benders_MICRO_ITERS.cpp
ze-hedi: if the parse succeeded it should be an object no ? so it's not necessary to double check as this bloc of code in under parsefromstream condition

PR #1288, opened 2026-09-06
Adding single responsability class to handle cut creating and adding on Master.

ze-hedi:
# New design philosophy

This PR reshapes the Benders kernel around two ideas: inheritance should express exactly one thing, and `BendersBase` should coordinate work rather than perform it.

## Where we are today

The kernel currently has three implementations. `BendersSequential` inherits from `BendersBase`, `BendersMPI` inherits from `BendersBase`, and `BendersByBatch` inherits from `BendersMPI`.
benders_kernel_class_hierarchy

`BendersBase` is the parent class and the centre of gravity. It owns the master problem and the subproblem map, the cut computation and aggregation, the bound and gap bookkeeping, the stopping criteria, option handling, timers, and everything I/O-shaped: console logs, the iteration trace, JSON output, and writing the final solution. Its subclasses call into it; it does not call much back.

`BendersSequential` is the simplest workflow. One process runs the whole loop: solve the master, walk every subproblem in turn, collect each cut, update the bounds, check convergence. No communication layer, so it also serves as the reference implementation the others are compared against.

`BendersMPI` runs the same loop, distributed. Rank 0 solves the master and broadcasts the trial investment vector, each rank solves only the subproblems it owns, and the cuts and aggregated bound come back through collective calls before the master is re-solved. The algorithm is unchanged; what it adds is partitioning, collectives, and synchronisation points.

`BendersByBatch` is a different iteration policy layered on the MPI one. Instead of solving every subproblem before touching the master, it splits them into batches and processes batches until the accumulated information already justifies re-solving the master, then stops early. It reuses the MPI communication substrate and overrides the loop that decides how much work an iteration does.

## Why this needs to change

None of the above is the result of a bad decision. It is the result of many locally reasonable ones. Each new capability was added by extending whatever class was closest to hand, and the structure we have is the accumulation of those extensions rather than a design anyone would choose today. That has three concrete costs.

**`BendersBase` is a god class.** Cut algebra, solver lifecycle, convergence policy, logging, serialisation, and instrumentation all live in one type. They change for entirely different reasons and at entirely different rates, but they share a header, a translation unit, and a mutable state. Touching cut computation means recompiling and re-reasoning about everything else.

### `BendersBase` becomes an orchestrator

benders_architecture
Everything that is not the loop moves out into collaborators with a single responsibility each.

- **`BendersCutsManager`** manages the creation and assembly of Benders cuts from subproblem dual solutions, including computation of separation points (xcut), aggregation of costs, and broadcasting cut data across ranks.
- **`BendersSubProblemsManager`** handles creation, storage, distribution, and solving of subproblems across multiple cache modes, with support for parallel execution, basis management, and problem-to-id mapping.
- **`BendersMasterManager`** manages the master LP problem lifecycle including variable/constraint management, objective function updates, and feeding cuts to the solver.
- **`BendersOutputManager`** orchestrates logging, CSV trace output, iteration data conversion, and result serialization throughout the algorithm and at completion.
- **`BendersOuterLoopManager`** manages the outer loop execution including criterion computation, solution tracking, and coordination between Benders iterations and outer loop objectives.

`BendersBase` keeps only the loop: ask the subclass for its iteration, feed the results through the collaborators, check whether to continue. It coordinates; it no longer computes.

## What this buys us

**The algorithm becomes testable.** Cut construction, gap computation, and stopping rules can each be unit tested against a fake or a fixture, with no solver and no filesystem. Failures become specific instead of "the end-to-end run diverged".

**Adding a variant costs one class.** A new execution strategy is a new sibling that implements the iteration and reuses every collaborator unchanged. Today it would mean deciding which existing class to inherit from and inheriting whatever else came with it.

**Changes stay local.** Changing the trace format touches `BendersLogger`. Changing the aggregation rule touches `BendersCutsManager`. Neither forces the other to be re-read, re-reviewed, or re-tested.

**Collaborators are independently replaceable.** Because each one is reached through a narrow interface, alternative implementations (a different stopping rule, a different output format) become substitutions rather than edits to shared code.

**The code explains itself.** A newcomer can read `BendersBase` and see the algorithm as a sequence of named steps, then go one level deeper only into the step they care about.

## Scope of this PR

This PR delivers the first collaborator extraction: **`BendersCutsManager`**.

All cut-related logic — building a cut from a subproblem's dual solution, gathering cuts across ranks, aggregating subgradients, and handing the result to the master — has been pulled out of the Benders classes and into a dedicated family of cuts managers.

### What was done

1. **Introduced `BendersCutsManager<Derived>` as a CRTP base** (compile-time polymorphism, zero virtual dispatch overhead). It owns the shared cut algebra: `ComputeXCut()` for the separation point, `BuildAllAggregatedCuts()` for subgradient aggregation, and `SetSubproblemDataCostAndSimplexIter()` for cost/iteration bookkeeping.

2. **Three concrete implementations**, one per Benders variant, all siblings with no inheritance relationship between them:
   - `BendersCutsManagerSequential` — wraps local data and calls the shared aggregation directly. No communication.
   - `BendersCutsManagerMpi` — MPI-gathers subproblem data to rank 0, then rank 0 runs aggregation and adds cuts to the master.
   - `BendersCutsManagerByBatch` — same gather as MPI, plus computes per-batch gap contributions so the by-batch loop can decide whether to stop early.

3. **Unified the aggregation path.** Previously `WorkerMaster` had multiple cut-adding methods (`add_cut`, `add_dynamic_cut`, `add_cut_by_iter`, `addSubproblemCut`) that each Benders variant called differently. These were replaced by a single `addGroupSubproblemCut()` entry point, called uniformly from `BuildAllAggregatedCuts()`. ~250 lines removed from `WorkerMaster`.

4. **Adapted the `BendersBase` unit tests** to construct and exercise the cuts manager, and added a guard for the border case of zero subproblems.

### What is intentionally unchanged

- The class hierarchy itself (`BendersByBatch` still inherits from `BendersMPI`). Flattening the hierarchy is a later step; this PR only extracts the first collaborator.
- The iteration loop, convergence logic, logging, and output writing remain inside the existing classes.
- No behavioural change: the same cuts are built, aggregated, and added in the same order.

### Design choice worth noting

CRTP was chosen over virtual dispatch because the cuts manager is called in the inner loop of every iteration, the type is always known statically within each Benders variant, and this avoids an indirection on a hot path. Each variant holds its own typed member — `BendersBase` does not hold a base pointer.

## Review notes

Suggested reading order:
1. `BendersCutsManager.hxx` — the shared CRTP base with `ComputeXCut`, `BuildAllAggregatedCuts`
2. `BendersCutsManagerSequential.h` — simplest implementation
3. `BendersCutsManagerMpi.h` → `BendersCutsManagerByBatch.h` — the two MPI-aware implementations (siblings, shared gather pattern but no inheritance between them)
4. Diff on `WorkerMaster` — see the removed aggregation methods
5. Test changes in `BendersBase` unit tests


tbittar:
Should we add the design description proposal as an ADR ? @JasonMarechal25 @ze-hedi 
Also for the open question, I would choose to handle MPI as a collaborator, especially if we add some more Benders variants in the future, we could reuse it rather than copying it again

tests/cpp/benders/BendersCutsManagerTest.cpp
tbittar, line 1: The implementations of GatherAndBuildCutsImpl should also be unit tested

tbittar (commented):
The design is very clear to me !

JasonMarechal25 (approved):

PR #1299, opened 2026-09-16
Allow empty TS for candidates [ANT-5837]

payetvin:


tests/python/test_candidates_reader.py
```diff
@@ -0,0 +1,22 @@
...
+    profile = CandidatesReader._read_or_create_link_profile_array_simple(
+        str(profile_file)
+    )
+
+    assert profile.shape == (8760,)
+    assert np.array_equal(profile, np.zeros(8760))
+
+
+def test_missing_profile_uses_default_unity_timeseries():
+    profile = CandidatesReader._read_or_create_link_profile_array_simple("")
```
flomnes:
```suggestion
    profile = CandidatesReader._read_or_create_link_profile_array_simple()
```

src/python/antares_xpansion/candidates_reader.py
```diff
@@ -212,11 +212,15 @@ def check_nan_in_profile_link_array(link_profile_array, file: str):
 
     @staticmethod
     def _read_or_create_link_profile_array_simple(file: str):
```
flomnes: The empty string `""` is a special value. Special values are bad
```suggestion
    def _read_or_create_link_profile_array_simple(file: str | None = None):
```

src/python/antares_xpansion/candidates_reader.py
```diff
@@ -212,11 +212,15 @@ def check_nan_in_profile_link_array(link_profile_array, file: str):
...
-        link_profile_array = np.ones(8760)
-        if file:
-            link_profile_array = np.genfromtxt(file)
-            CandidatesReader.check_nan_in_profile_link_array(
-                link_profile_array, file)
+        if not file:
+            return np.ones(8760)
+
+        if not Path(file).read_text().strip():
+            return np.zeros(8760)
```
flomnes, line 220: Add comment here "the file is empty except for space, tab, etc."

src/cpp/lpnamer/input_reader/LinkProfileReader.cpp
```diff
@@ -1,9 +1,24 @@
...
 #include <filesystem>
+#include <fstream>
+#include <iterator>
 
 #include "antares-xpansion/xpansion_interfaces/LogUtils.h"
 
+namespace
+{
+bool isEmptyProfileFile(const std::filesystem::path& filename)
+{
```
flomnes: Better name
```suggestion
bool isFileEmptyOrSpaceOnly(const std::filesystem::path& filename)
{
```

src/cpp/lpnamer/input_reader/LinkProfileReader.cpp
```diff
@@ -13,9 +28,29 @@ std::vector<LinkProfile> LinkProfileReader::ReadLinkProfile(
...
+    const bool empty_direct_profile = isEmptyProfileFile(direct_filename);
+    const bool empty_indirect_profile = isEmptyProfileFile(indirect_file_name);
     std::vector<LinkProfile> result;
     ReadLinkProfile(direct_filename, result, true);
     ReadLinkProfile(indirect_file_name, result, false);
+
+    if (result.empty())
+    {
+        result.emplace_back(logger_);
+    }
```
flomnes, line 40: This block makes no sense to me

Why would we add a logger to a vector of `LinkProfile` ? Does this code even build ?

OK ignore this
```cpp
explicit LinkProfile(std::shared_ptr<ProblemGenerationLog::ProblemGenerationLogger> logger):
        logger_(logger)
    {
    }
```

src/cpp/lpnamer/input_reader/LinkProfileReader.cpp
```diff
@@ -13,9 +28,29 @@ std::vector<LinkProfile> LinkProfileReader::ReadLinkProfile(
...
+        {
+            std::fill(profile.direct_link_profile.begin(), profile.direct_link_profile.end(), 0);
+        }
+        if (empty_indirect_profile)
+        {
+            std::fill(profile.indirect_link_profile.begin(),
+                      profile.indirect_link_profile.end(),
+                      0);
+        }
+    }
```
flomnes, line 53: There are 3 `ReadLinkProfile` methods
- A 2-argument method, which I'll call "2"
- A 3-argument method, which I'll call "3"
- A 1-argument method, which I'll call "1"

The call graph looks like (-> means "calls")
2 -> 3
1 -> 3
3 doesn't call any of 1, 2, it's the "low-level" method

Then the higher-level API function `LinkProfileReader::importProfile` calls 2

So if we build the complete chain

`LinkProfileReader::importProfile` -> 2 -> 3

1 is never called (probably dead code) so you don't need to update it in fact I'd encourage you to remove it for clarity

Then choose one of 2/3 and update it, but only one.

flomnes (changes requested):

tbittar (commented):
All good once Florian's comments are fixed

src/python/antares_xpansion/input_checker.py
```diff
@@ -49,13 +49,16 @@ def _check_profile_file_consistency(filename_path):
...
-            if first_profile[-1] < 0:
+            if first_profile and first_profile[-1] < 0:
                 logger.error(
                     "Line %d in file %s indicates a negative value"
                     % (idx + 1, filename_path)
                 )
                 raise ProfileFileNegativeValue
 
+    if not first_profile:
+        return False
```
flomnes, line 51: Suggestion - use [numpy.loadtxt](https://numpy.org/doc/stable/reference/generated/numpy.loadtxt.html#numpy.loadtxt) to load the CSV file, then validate it's contents

Don't re-invent the wheel

flomnes (approved):

PR #1300, opened 2026-09-18
ANT-5787 Standardize xpansion classic and trajectory

YoshuaLisasi:
**Unify CLI options and clean up trajectory launcher**

### Context

Several inconsistencies in the CLI were reported:

The classic launcher uses --problem-format while the trajectory launcher uses --problems-format
--trajectory does not appear in antares-xpansion-launcher --help
launch_trajectory.py is unused since the unified launcher in launch.py

### Changes

Rename --problems-format to --problem-format in args_parser_trajectory.py to align with the classic launcher
Add --trajectory to InputParser in input_parser.py so it appears in --help (the flag is still intercepted early in launch.py before argparse runs)
Remove launch_trajectory.py — dead code, no references anywhere in the project

How to test : 
```
python src/python/launch.py --help              # --trajectory should appear
python src/python/launch.py --trajectory --help  # trajectory-specific help should display
```


JasonMarechal25 (approved):

src/python/antares_xpansion/trajectory/args_parser_trajectory.py
```diff
@@ -106,7 +106,7 @@ def _initialize_parser(self):
         )
         # When manipulating problem files : under which format are they written ?
         self.parser.add_argument(
-            "--problems-format",
+            "--problem-format",
             dest=TrajectoryLauncherOptionsKeys.problems_format_key(),
             type=str,
```
flomnes: This returns `PROBLEMS_FORMAT`, which is probably not what we want. I suggest renaming all `problems_format` to `problem_format` in the code.

grep returns 64 references on `problems_format` on your branch. Should be 0

flomnes (changes requested):

src/python/antares_xpansion/trajectory/args_parser_trajectory.py
JasonMarechal25: Good thinking. I wasn't so throughful

